PR #42, opened 2014-11-07
Fix for build issues when the job description has "invalid" characters.

Lakritzator:
I had some problems with the StashNotifier. On some builds the INPROGRESS worked but the SUCCESSFUL not. After looking into the differences, the thing I found was that the jobs which weren't placed on SUCCESSFUL had a custom build description which among others had "/" in it.

This change should fix it!

(somehow I can't use the plugin when I build it myself: class org.jenkinsci.plugins.stashNotifier.StashNotifier is missing its descriptor)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -627,7 +627,7 @@ public class StashNotifier extends Notifier {
                 replaceAll("\\\\u00BB", "\\/");
         json.put("name", fullName);
 
-        json.put("description", getBuildDescription(build, state));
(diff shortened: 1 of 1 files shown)


gruetter:
Het @Lakritzator   Did you try and see if your last proposal solves your problem?

Lakritzator:
Oh yeah, so much things on my mind that I forgot about this!
Thanks for reminding me!

I can confirm that the problem is certainly with the encoding! The "umlaut" caused the problem, changing the build description from haven "für" to something without the umlaut made the notify work again.

Due to a couple of hospital visits (nothing bad), and holidays, I haven't had time to try build the change.
As I had problems deploying the plugin, after building it myself, I was somewhat unmotivated to continue to work.

I will have another go at building the change.

Lakritzator:
My change seems to work, will create a new pull-request so we don't confuse the masses :)

Issue #46, opened 2015-01-13
HTTPClient and SNI support. Causes StashNotifier to fail on vaild SSL cert


nightwolfzor:
I notice that we are still using HttpClient 4.1 which doesnt support SNI. Which means that if stash is running on the same server as another SSL site we get problems. 

See https://issues.apache.org/jira/browse/HTTPCLIENT-1119

We need to upgrade to 4.3.2+. This would be good to do anyway considering the number of security holes fixed since 4.1 anyway.

PR #47, opened 2015-01-13
Fix for issue #46 - allow SNI and update HttpClient to version 4.3.5

nightwolfzor:
My attempted to move to HttpClient 4.3.x which gives SNI to stop SSL errors (even when unsecure is used) because the sever IP is hosting multiple HTTPS domains. 

Fixes issue #46


gruetter:
Hey @nightwolfzor . Thanks a bunch for your PR. Much appreciated.

PR #48, opened 2015-01-14
Notify with commit that triggered the build with using "Merge before build".

klieber:
Currently, if you use the "Merge before build" option in the Git plugin then the stash notifier plugin send the commit id of the merge to stash which means that the pull request never gets the build status.  This change just sends both the merge commit and the original commit to Stash.

I had to modify the version of Jenkins and the Git plugin in order to get this to work as I expected.  Not sure if that is a problem or not.  Also, it might make more sense to make this a configurable option but this is my first time working on a Jenkins plugin so I didn't get that far into it at this point.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,9 +55,9 @@
 			<version>4.1</version>
 		</dependency>
 		<dependency>
-			<groupId>org.jenkinsci.plugins</groupId>
+			<groupId>org.jenkins-ci.plugins</groupId>
 			<artifactId>git</artifactId>
-			<version>1.1.26</version>
+			<version>2.0</version>
 			<type>jar</type>
 		</dependency>
 	</dependencies>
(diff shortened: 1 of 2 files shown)


gruetter:
Hey Kyle.  Thanks a bunch for your contribution. As much as I'd like to merge it to develop, I fear the dependency to the very recent Jenkins version will break the plugin for many users out there. The only solution I see is to temporarily maintain a dedicated branch for users that require the admittedly handy pre merge feature. What do you think, @klieber ?

klieber:
@gruetter I looked into this further and it looks like I didn't need to require a newer Jenkins version. All I needed was to require the 2.0 version or higher of the Git Plugin.  I updated my branch with that change.

gruetter:
thanks for following up on this, @klieber . Looks good to me.

PR #49, opened 2015-01-14
Including an extra field to set custom cookies

martoc:
SSO does not need user and password credentials, In our case we pass the token in a cookie, a new field was added to the advance view and if the cookie is set user and password fields are optional


scaytrase:
@martoc any updates planned?

scaytrase:
Closing as no response from author received

PR #51, opened 2015-02-12
[JENKINS-24967] Fix proxy exclusions not working

### LINKED ISSUE #41: Plugin does not seem to use Jenkins no-proxy settings
If proxy has been setup under "Manage Jenkins -> Manage Plugins -> Advanced" the StashNotifier will always try to access the Stash instance through proxy even though it matches the "No Proxy Hosts" wildcard set in the configuration.

https://issues.jenkins-ci.org/browse/JENKINS-24967

bbaetz:
Also support proxy authentication

Fixes #41

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -41,23 +41,15 @@ import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.config.Registry;
 import org.apache.http.config.RegistryBuilder;
-import org.apache.http.conn.ClientConnectionManager;
 import org.apache.http.conn.HttpClientConnectionManager;
-import org.apache.http.conn.params.ConnRoutePNames;
-import org.apache.http.conn.scheme.PlainSocketFactory;
-import org.apache.http.conn.scheme.Scheme;
-import org.apache.http.conn.scheme.SchemeRegistry;
 import org.apache.http.conn.socket.ConnectionSocketFactory;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
 import org.apache.http.conn.ssl.SSLContexts;
-import org.apache.http.conn.ssl.SSLSocketFactory;
 import org.apache.http.conn.ssl.TrustStrategy;
 import org.apache.http.entity.StringEntity;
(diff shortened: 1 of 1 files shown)


bbaetz:
I've tested this on a local Jenkins instance with a local proxy in a number of scenarios (including wildcards) and it seems to work, but some extra testing would be good.

I didn't try too hard to work out what was wrong with the original code - I figured that letting core Jenkins manage the logic was simpler anyway.

gruetter:
Thanks a bunch, @bbaetz. I'll be alle to Check this Out tomorrow. I meant to tackle this myself but just can't seem to find the time.

gruetter:
Ok, I did have a look at the code today after all ;-). I actually wasn't able to compile it using Maven. The reason is that the current parent pom of this plugin still requires a java language level of 1.5 (meh) and you used `InetSocketAddress.getHostString()`, which has been added with Java 1.7. I guess you did not compile this with maven, did you @bbaetz ? We could use `InetSocketAddress.getHostName()` instead. Not sure the involved name service reverse lookup could pose a problem here. What do you think, @bbaetz ?

Other than that, do I understand the code right, that the no-proxy configuration is taken from the Jenkins proxy configuration and that the built `HttpClient`, configured based on this proxy configuration, will honor the no-proxy configuration?

bbaetz:
Nope, I used mvn. But the compiler config only affects the language version not the library version, and I was compiling with 1.8. Will have to look for another option. And find a workable jdk5 install....

Don't want to force a reverse lookup if I can avoid it, in case there's a private IP without proper DNS.

I wish there weren't so many different and incompatible ways of dealing with http connections...

Yes, that's the idea.

bbaetz:
OK, try this. I decided to grab the proxy details from the Proxy rather than getting it again from the Jenkins ProxyConfiguration to avoid doing another DNS lookup.

If Jenkins ever supports NTLM proxy auth that won't work (the hostname is needed) but:

a) Jenkins doesn't support that anyway (https://issues.jenkins-ci.org/browse/JENKINS-3350)
b) This is still better than the current situation before the patch of not supporting authentication at all
c) By the time (a) is fixed, Jenkins might stop supporting a Java version that was EOL in 2009

gruetter:
Thanks for following up, @bbaetz ! Your solution looks good to me.  I'll run some tests later today and will merge.

gruetter:
Works like a charm on my local machine. Thanks a bunch, Bradley.

PR #55, opened 2015-04-27
Abbreviate field values if needed

lukasniemeier-zalando:
The Stash rest endpoint for notifying the build status has some restrictions regarding the length of transmitted fields (see https://developer.atlassian.com/static/rest/stash/3.6.0/stash-build-integration-rest.html). This PR ensures that the fields comply to this (strange) contract.


gruetter:
thanks a buch @lukasniemeier-zalando for this PR! Nice addition.

PR #58, opened 2015-05-10
Feature/npe

magnusr:
When all builds for a job so far hav been unable to clone, the last built revision may be null. This change attempts to handle this by skipping that BuildData. Also adds unit tests for the normal (non-null BuildData) and abnormal cases.


gruetter:
Hey @magnusr - thanks for this PR. Much appreciated. I have never got around to try Mockito but I do see it's power now.

PR #60, opened 2015-05-19
Enhanced project keys

### LINKED ISSUE #36: Please provide per Jenkins-Project Key setting
I would like to get the ability the change the KEY-json field which is sent to the Stash server on a per project basis. Basically i want to be able to set the same Key for several Jenkins projects which map to the same Stash repoitory.
The Key should default to the Project name, but be editable.

scaytrase:
Here is implementation for two options 

* Custom persist project key for a job (Fix #36 )
* Option to prepend parent project name for nested project (i.e maven subprojects). This helps to different jobs operating subprojects with same name to keep different notifications instead of overriding each other.

All changes should be BC as disabled on empty (by default). All changes built and tested on my own jenkins installation.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -94,22 +94,28 @@ public class StashNotifier extends Notifier {
 
 	/** base url of Stash server, e. g. <tt>http://localhost:7990</tt>. */
 	private final String stashServerBaseUrl;
-	
+
 	/** name of Stash user for authentication with Stash build API. */
 	private final String stashUserName;
-	
+
 	/** password of Stash user for authentication with Stash build API. */
 	private final Secret stashUserPassword;
-	
+
 	/** if true, ignore exception thrown in case of an unverified SSL peer. */
 	private final boolean ignoreUnverifiedSSLPeer;
 
 	/** specify the commit from config */
 	private final String commitSha1;
-	
+
 	/** if true, the build number is included in the Stash notification. */
 	private final boolean includeBuildNumberInKey;
-	
+
+	/** specify project key manually */
+	private final String projectKey;
+
+	/** append parent project key to key formation */
+	private final boolean prependParentProjectKey;
+
 	// public members ----------------------------------------------------------
 
 	public BuildStepMonitor getRequiredMonitorService() {
@@ -123,7 +129,9 @@ public class StashNotifier extends Notifier {
 			String stashUserPassword,
 			boolean ignoreUnverifiedSSLPeer,
 			String commitSha1,
-			boolean includeBuildNumberInKey) {
+			boolean includeBuildNumberInKey,
+			String projectKey,
+			boolean prependParentProjectKey) {
 		this.stashServerBaseUrl = stashServerBaseUrl.endsWith("/")
                 ? stashServerBaseUrl.substring(0, stashServerBaseUrl.length()-1)
                 : stashServerBaseUrl;
@@ -133,6 +141,8 @@ public class StashNotifier extends Notifier {
 			= ignoreUnverifiedSSLPeer;
 		this.commitSha1 = commitSha1;
 		this.includeBuildNumberInKey = includeBuildNumberInKey;
+		this.projectKey = projectKey;
+		this.prependParentProjectKey = prependParentProjectKey;
 	}
 
 	public String getStashServerBaseUrl() {
@@ -146,11 +156,11 @@ public class StashNotifier extends Notifier {
 	public String getStashUserPassword() {
 		return stashUserPassword.getEncryptedValue();
 	}
-	
+
 	public boolean getIgnoreUnverifiedSSLPeer() {
 		return ignoreUnverifiedSSLPeer;
 	}
-	
+
 	public String getCommitSha1() {
 		return commitSha1;
 	}
@@ -158,19 +168,27 @@ public class StashNotifier extends Notifier {
 	public boolean getIncludeBuildNumberInKey() {
 		return includeBuildNumberInKey;
 	}
-	
-	@Override
+
+    public String getProjectKey() {
+        return projectKey;
+    }
+
+    public boolean getPrependParentProjectKey() {
+        return prependParentProjectKey;
+    }
+
+    @Override
 	public boolean prebuild(AbstractBuild<?, ?> build, BuildListener listener) {
 		return processJenkinsEvent(build, listener, StashBuildState.INPROGRESS);
 	}
-	
+
 	@Override
 	public boolean perform(
-			AbstractBuild<?, ?> build, 
-			Launcher launcher, 
+			AbstractBuild<?, ?> build,
+			Launcher launcher,
 			BuildListener listener) {
-		
-		if ((build.getResult() == null) 
+
+		if ((build.getResult() == null)
 				|| (!build.getResult().equals(Result.SUCCESS))) {
 			return processJenkinsEvent(
 					build, listener, StashBuildState.FAILED);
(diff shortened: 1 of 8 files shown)


gruetter:
Hey @scaytrase - thanks a bunch for your PR! Works great here. I'll review the code and merge later today.

scaytrase:
@gruetter Ok, tell me if I can do something more here

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -636,21 +672,50 @@ private HttpEntity newStashBuildNotificationEntity(
...
+
+		if (overriddenKey != null && overriddenKey.trim().length() > 0) {
+			PrintStream logger = listener.getLogger();
+			try {
+				EnvVars environment = build.getEnvironment(listener);
+				return StringEscapeUtils.escapeJavaScript(environment.expand(projectKey));
+			} catch (IOException e) {
+				logger.println("Unable to expand commit SHA value");
+				e.printStackTrace(logger);
+				return null;
```
gruetter: if you return `null` here, you would set `null` as the value of the JSON key `key` in line 662 above, which doesn't check for a null value returned by this method. Instead, I propose to log an error message like you do but instead use the project name as a default key.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -636,21 +672,50 @@ private HttpEntity newStashBuildNotificationEntity(
...
+			} catch (InterruptedException e) {
+				logger.println("Unable to expand commit SHA value");
+				e.printStackTrace(logger);
+				return null;
```
gruetter: same issue as mentioned [above](https://github.com/jenkinsci/stashnotifier-plugin/pull/60/files#r30740915)

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -636,21 +672,50 @@ private HttpEntity newStashBuildNotificationEntity(
...
+	private String getBuildKey(final AbstractBuild<?, ?> build,
+							   BuildListener listener) {
+
+        String overriddenKey = (projectKey != null && projectKey.trim().length() > 0) ? projectKey : getDescriptor().getProjectKey();
+
+		if (overriddenKey != null && overriddenKey.trim().length() > 0) {
+			PrintStream logger = listener.getLogger();
+			try {
+				EnvVars environment = build.getEnvironment(listener);
+				return StringEscapeUtils.escapeJavaScript(environment.expand(projectKey));
```
gruetter: By returning from here, you effectively prevent the evaluation of  `prependParentProjectKey` and `includeBuildNumberInKey` settings. While I think this makes sense, I propose to add a corresponding remark to the tooltip of the field _Override project key_ to avoid confusion.

src/main/resources/org/jenkinsci/plugins/stashNotifier/StashNotifier/help-prependParentProjectKey.html
```diff
@@ -0,0 +1,3 @@
+<div>
+  Whether to prepend parent project while forming notify key
```
gruetter: I propose to add a comment that the parent project name is not prepended if the user specified an override key as per [this remark above](https://github.com/jenkinsci/stashnotifier-plugin/pull/60/files#r30741270).

src/main/resources/org/jenkinsci/plugins/stashNotifier/StashNotifier/help-projectKey.html
```diff
@@ -0,0 +1,3 @@
+<div>
+  If not empty, this key will be used instead of automatically configured project key
```
gruetter: I propose to mention that the value of the setting _Prepend parent project key_ is ignored if an override project key is specified.

src/main/resources/org/jenkinsci/plugins/stashNotifier/StashNotifier/global.jelly
```diff
@@ -25,5 +25,10 @@
                help="${rootURL}/plugin/stashNotifier/help-globalConfig-includeBuildNumberInKey.html">
           <f:checkbox />
       </f:entry>
+      <f:entry title="Prepend parent project key"
```
gruetter: Just as an idea to clarify: how about we rename the field label to "Prepend parent project name to key"?

gruetter:
Finished the review, @scaytrase . Besides the `null` issue, all minor findings.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
scaytrase: I also see some forgotten copy-paste by me from here. I'll fix that.

src/main/resources/org/jenkinsci/plugins/stashNotifier/StashNotifier/help-prependParentProjectKey.html
scaytrase: I'm not sure if we should really try to prepend parent key instead.

scaytrase:
Thanks,  @gruetter  I'll try to fix all today or at least tomorrow morning

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -636,21 +672,50 @@ private HttpEntity newStashBuildNotificationEntity(
...
+	 *
 	 * @param 	build	the build to notify Stash of
 	 * @return	the build key for the Stash notification
 	 */
-	private String getBuildKey(final AbstractBuild<?, ?> build) {
-		StringBuilder key = new StringBuilder();
+	private String getBuildKey(final AbstractBuild<?, ?> build,
+							   BuildListener listener) {
+
+        String overriddenKey = (projectKey != null && projectKey.trim().length() > 0) ? projectKey : getDescriptor().getProjectKey();
```
scaytrase: I'm not sure it is the best way to check that parameter is not empty. Is this correct? `projectKey` is never set up at global configuration, so I can miss something due to I'm new to Jenkins API

scaytrase:
well, cloudbees automerge now fails. should I rebase or somethink like?

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -636,51 +672,93 @@ private HttpEntity newStashBuildNotificationEntity(
...
+
+		StringBuilder key = new StringBuilder();
+
+		if (prependParentProjectKey || getDescriptor().isPrependParentProjectKey()){
+			if (null != build.getParent().getParent()) {
+				key.append(build.getParent().getParent().getFullName()).append('-');
+			}
+		}
+
+		String overriddenKey = (projectKey != null && projectKey.trim().length() > 0) ? projectKey : getDescriptor().getProjectKey();
```
scaytrase: I'm not sure it is the best way to check that parameter is not empty. Is this correct? projectKey is never set up at global configuration, so I can miss something due to I'm new to Jenkins API

scaytrase:
@gruetter Rebased

src/main/resources/org/jenkinsci/plugins/stashNotifier/StashNotifier/help-prependParentProjectKey.html
gruetter: I like that better as well. Good call.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
gruetter: I'm not much of a Jenkins pro either but your code looks good to me. I used the same approach and it has worked just fine in the past.

gruetter:
Looks all good to me know. I'll merge. Thanks for following up @scaytrase . Great contribution.

scaytrase:
Thank you for good plugin we can contribute to! :thumbsup:

Issue #56, opened 2015-04-27
Use credential plugin


sean-abbott:
I'd like this plugins to use the credentials plugin, because, for instance, I'm being required to provide a client certificate and your plugin doesn't have a way to do that.

adou600:
Me and my team would be also very interested into that feature!

matthewbarr:
Has anyone tried to do this yet?  We're using the token plugin for now, but also have a cert based authentication that is much more compliant w/ our general security protocols.

SchwaFa:
@matthewbarr Which token plugin do you use?

PR #62, opened 2015-06-15
Integration of the credentials plugin

pdodds:
I have had a go at integrating the credentials plugin into the stash notifier.  This allows you to use either a username/password or a certificate for security.

I believe this is referenced in issue #56

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,18 @@
 			<type>jar</type>
 		</dependency>
 		<dependency>
+			<groupId>org.jenkins-ci.plugins</groupId>
+			<artifactId>credentials</artifactId>
+			<version>1.22</version>
+			<type>jar</type>
+		</dependency>
+		<dependency>
+			<groupId>org.jenkins-ci.plugins</groupId>
+			<artifactId>plain-credentials</artifactId>
+			<version>1.1</version>
+			<type>jar</type>
+		</dependency>
+		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
 			<version>4.12</version>
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashCredentialMatcher.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashCredentialMatcher.java
@@ -1,0 +1,15 @@
+package org.jenkinsci.plugins.stashNotifier;
+
+import com.cloudbees.plugins.credentials.Credentials;
+import com.cloudbees.plugins.credentials.CredentialsMatcher;
+import com.cloudbees.plugins.credentials.common.CertificateCredentials;
+import com.cloudbees.plugins.credentials.common.UsernamePasswordCredentials;
+
+/**
+ * A very simple matcher to ensure we only show username/password or certificate credentials
+ */
+public class StashCredentialMatcher implements CredentialsMatcher {
+    public boolean matches(Credentials credentials) {
+        return (credentials instanceof CertificateCredentials) || (credentials instanceof UsernamePasswordCredentials);
+    }
+}
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -15,28 +15,33 @@
  */
  package org.jenkinsci.plugins.stashNotifier;
 
+import com.cloudbees.plugins.credentials.Credentials;
+import com.cloudbees.plugins.credentials.CredentialsMatchers;
+import com.cloudbees.plugins.credentials.CredentialsProvider;
+import com.cloudbees.plugins.credentials.common.*;
+import com.cloudbees.plugins.credentials.domains.DomainRequirement;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.Launcher;
 import hudson.ProxyConfiguration;
-import hudson.model.AbstractBuild;
-import hudson.model.AbstractProject;
-import hudson.model.BuildListener;
-import hudson.model.Result;
-import hudson.plugins.git.util.BuildData;
+import hudson.model.*;
 import hudson.plugins.git.Revision;
+import hudson.plugins.git.util.BuildData;
+import hudson.security.ACL;
+import hudson.security.AccessControlled;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.BuildStepMonitor;
 import hudson.tasks.Notifier;
 import hudson.tasks.Publisher;
 import hudson.util.FormValidation;
-import hudson.util.Secret;
+import hudson.util.ListBoxModel;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
 import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHost;
 import org.apache.http.HttpResponse;
+import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpPost;
@@ -45,13 +50,17 @@ import org.apache.http.config.RegistryBuilder;
 import org.apache.http.conn.HttpClientConnectionManager;
 import org.apache.http.conn.socket.ConnectionSocketFactory;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
+import org.apache.http.conn.ssl.SSLContextBuilder;
 import org.apache.http.conn.ssl.SSLContexts;
 import org.apache.http.conn.ssl.TrustStrategy;
 import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.auth.BasicScheme;
+import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.http.impl.client.ProxyAuthenticationStrategy;
 import org.apache.http.impl.conn.BasicHttpClientConnectionManager;
 import org.apache.http.util.EntityUtils;
+import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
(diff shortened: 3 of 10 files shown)


gruetter:
Hey @pdodds . I just tried out your changes locally. I like the way you integrated the credentials plugin from a users perspective - nice. That said, I wasn't able to get my test build to notify stash successfully. The console log showed the following exception

    Caught exception while notifying Stash with id 0934773e8e6895191b736fec1c83768f44e8d74a
    java.lang.NullPointerException
	    at com.cloudbees.plugins.credentials.matchers.IdMatcher.<init>(IdMatcher.java:49)
	    at com.cloudbees.plugins.credentials.CredentialsMatchers.withId(CredentialsMatchers.java:112)
	    at org.jenkinsci.plugins.stashNotifier.StashNotifier.createRequest(StashNotifier.java:624)
	    at org.jenkinsci.plugins.stashNotifier.StashNotifier.notifyStash(StashNotifier.java:585)
	    at org.jenkinsci.plugins.stashNotifier.StashNotifier.processJenkinsEvent(StashNotifier.java:227)
	    at org.jenkinsci.plugins.stashNotifier.StashNotifier.prebuild(StashNotifier.java:180)
	    at hudson.model.AbstractBuild$AbstractBuildExecution.preBuild(AbstractBuild.java:804)
	    at hudson.model.AbstractBuild$AbstractBuildExecution.preBuild(AbstractBuild.java:799)
	    at hudson.model.Build$BuildExecution.doRun(Build.java:142)
	    at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:533)
	    at hudson.model.Run.execute(Run.java:1754)
	    at hudson.model.FreeStyleBuild.run(FreeStyleBuild.java:43)
	    at hudson.model.ResourceController.execute(ResourceController.java:89)
	    at hudson.model.Executor.run(Executor.java:240)

so I guess `descriptor.credentials` was null which was not checked. I also see in my job configuration, that, even though I selected a credential in the _Credentials_ drop down box (the same credential which successfully works in the _Source Code Management_/_git_ section above), the error message _Please specify the credentials to use_ remains in place. The problem disappeared, as soon as I selected credentials in the global Jenkins configuration. 

So it seems the per-job credentials configuration is not working, yet. I haven't had time to actually check the code itself - its a bit crazy at the office right now 8-(. I'll try to find time later this week.

At any rate, thanks for tackling this long overdue feature addition! Much appreciated.

pdodds:
Sorry I missed this note - I'll get on it today and retest :)

pdodds:
I just have one more issue - i believe the UI is working - just need another day to get the tests finished :) i'm new to working with Jenkins so I think I capably made every mistake imaginable

gruetter:
> I just have one more issue - i believe the UI is working - just need another day to get the tests finished :) i'm new to working with Jenkins so I think I capably made every mistake imaginable
> 
Same here, really ;-). No worries

pdodds:
OK - I just retested it locally with a Stash instance and I believe the UI is all working the user/password stuff is back in shape.

pdodds:
Also I was thinking since you need to re-enter your credentials it might be better to do a major version bump.   I spoke to a another plugin dev and that seemed to be the suggestion when going to the credentials plugin?

gruetter:
I also think that makes sense. Not the least because we now have additional plugin dependencies.

gruetter:
I have just checked out the plugin after your latest changes, @pdodds . I'm still having problems with the global vs. local configuration. When I select a (valid) credential in the global settings and do not specify one (= select _none_) in the job configuration, the authentication fails. When using the same credentials in the job configuration, everything works fine. Looks to me as if the global config is not used at all. Can you confirm the same behaviour on your end, @pdodds ?

I've checked the code and have pushed one little change. However, I wasn't able to pinpoint the problem in the time I had. We should probably check out how other plugins solve this problem. Did you by any chance reuse the code from another project that we could check out?

gruetter:
I'm wondering: when you have a list box as an input field for a form, that always has a value, right? So if I select "-none-", that's the value it has and it always overrides the global configuration. If that's the problem, maybe using a global configuration with another field name would help. What do you think, @pdodds ?

pdodds:
Let me have a look again at the global configuration,  maybe it is a problem when applied there.   Actually I took the code from for the credentials by hunting around several plugins,  the documentation on how to use it is very high-level.

pdodds:
I forgot to check for global/local - actually more like I had no idea that was how you did it :) anyway,  I just retested with a global id set and no id set on the local.

SchwaFa:
When is this going to be merged? I'm really looking forward to have that feature! ;)

scaytrase:
@pdodds Could we deal with notification when local credentials are not selected?
image

I see two ways - disable requirement checking for local  configuration or just use a checkbox like `override global credentials`

`-none-` local credentials selected works well, but I think notifications will confuse users.

pdodds:
Sorry I'm out on paternity right now,  I'll try and get back to it in a week or two,  though if anyone can pick it up that would be brilliant :)

scaytrase:
@pdodds I'll try to put all local connection config under [optional block](https://wiki.jenkins-ci.org/display/JENKINS/Jelly+form+controls)

scaytrase:
It seems that optional block does not looks better than advanced button. I'll merge this as is.

scaytrase:
:+1: Thank you, @pdodds !

vassilevsky:
Hello :) Thank you for your contribution! 💪This is definitely the right way™ to store credentials.

However, the way I found out about it is that I got a bug report from one of the users of the Jenkins server that I set up. He was seeing this in the build log:

```
Caught exception while notifying Stash with id 383c81d6cd6a67372857726b7aa816c6df7b9de1
java.lang.NullPointerException
    at com.cloudbees.plugins.credentials.matchers.IdMatcher.<init>(IdMatcher.java:49)
    at com.cloudbees.plugins.credentials.CredentialsMatchers.withId(CredentialsMatchers.java:112)
    at org.jenkinsci.plugins.stashNotifier.StashNotifier.getHttpClient(StashNotifier.java:344)
    at org.jenkinsci.plugins.stashNotifier.StashNotifier.notifyStash(StashNotifier.java:621)
    at org.jenkinsci.plugins.stashNotifier.StashNotifier.processJenkinsEvent(StashNotifier.java:244)
    at org.jenkinsci.plugins.stashNotifier.StashNotifier.perform(StashNotifier.java:210)
    at hudson.tasks.BuildStepMonitor$1.perform(BuildStepMonitor.java:20)
    at hudson.model.AbstractBuild$AbstractBuildExecution.perform(AbstractBuild.java:785)
    at hudson.model.AbstractBuild$AbstractBuildExecution.performAllBuildSteps(AbstractBuild.java:726)
    at hudson.model.Build$BuildExecution.post2(Build.java:185)
    at hudson.model.AbstractBuild$AbstractBuildExecution.post(AbstractBuild.java:671)
    at hudson.model.Run.execute(Run.java:1766)
    at hudson.model.FreeStyleBuild.run(FreeStyleBuild.java:43)
    at hudson.model.ResourceController.execute(ResourceController.java:98)
    at hudson.model.Executor.run(Executor.java:410)
```

When I went in to investigate, I saw a warning on Manage Jenkins page that said something like "Your data is unreadable or corrupted" and there was this error for every existing job there:

> MissingFieldException: No field 'stashUserName' found in class 'org.jenkinsci.plugins.stashNotifier.StashNotifier', MissingFieldException: No field 'stashUserPassword' found in class 'org.jenkinsci.plugins.stashNotifier.StashNotifier'

Apparently someone has upgraded all plugins and this plugin upgrade completely borked the settings and notifications stopped working for all builds :)

I'd like to make it clear that this is not a complaint. I already have created the credentials the right way™ and it is working again.

I was just wondering if there was / is any way to migrate old settings upon plugin upgrade. So that notifications continue working after the upgrade.

Another issue is versioning. If there is no way to upgrade the plugin without breaking the installation, then, according to semantic versioning principles, it would have been better to release this as version 2.x.

Anyway, thanks again for this improvement. Please keep in mind that this information is not to belittle it, but to supplement :)

scaytrase:
Thank you for notifying. The only way I found now is to [mark plugin version as BC breakage](https://wiki.jenkins-ci.org/display/JENKINS/Marking+a+new+plugin+version+as+incompatible+with+older+versions)

But I think this is not a data migration. As of the credential support - I do not see a way to migrate credentials at all

ashnazg:
Was it intentional that this plugin's config dialogue does not recognize any SSH credentials that are saved globally?  I've found that only global credential profiles that are "Username and password" are available on the per-job dialogue... none of my SSH credential profiles are shown on the drop-down.

Even if I use the "Add" button on the stashnotifier config on a given job, input a new SSH credential profile and save it (I know it succeeds, because I later see it on the global credentials list), this plugin's Credentials drop-down will not show it.  It is only showing Username/Password credential profiles.

For future googlers:  if you've upgraded the stashnotifier plugin to 1.9.x, and previously used global SSH profiles for this plugin on your Jenkins jobs, you'll begin to see the NPEs that are referenced by Issue #71.  And since you cannot choose one of your globally saved SSH profiles, you will have to create a global username/password profile and specifically select it on each of your jenkins jobs.

scaytrase:
What is meant with `SSH credentials`? SSH login with private key? I cannot imagine using this type of credentials with HTTP REST API

ashnazg:
Ok, after downgrading back to 1.8.0 to see why it used to work fine, I now realize that stashnotifier itself had a global config, and it is there where I had a username/password saved.  So then, with the upgrade to 1.9.0, I need to create that same username/password as a Credentials plugin user profile.

I reckon I expected that all the global Credentials profiles I could see should have been available as options on the stashnotifier config drop-down.  Thus, the answer to my "was it intentional" question is a "YES" :)

scaytrase:
:+1:

ashnazg:
Also, after creating that one username/password global Credential profile, I will need to then update the stashnotifier config _on *each* jenkins job_ to select that global credential... right?

scaytrase:
Not think so, I'll recheck the source now, be patient

scaytrase:
Source says it [should fallback to global credentials](https://github.com/jenkinsci/stashnotifier-plugin/blob/develop/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java#L350-L353) if not configured. Feel free to report an issue otherwise

ashnazg:
Ah, just found that the global config for stashnotifier was still available...

ashnazg:
All good now... thanks @scaytrase!

PR #64, opened 2015-07-01
Allow cleanup 'old' builds in Stash after successful build

laurentverbruggen:
Stash has an issue that if you've set a requirement of minimum 1 successful build for pull requests to merge, you won't be able to merge it when a it contains a failed build. Even if the last build was successful!

You can mitigate this problem by either pushing a new commit in the pull request, causing the builds to 'disappear' since you've changed the commit, or change the statuses of those old builds to success (unfortunately Stash REST API doesn't allow you to remove them). The latter is what I included in this pull request.

You could argue that you can solve this by not enabling the feature 'Keep repeated builds in Stash', but this won't solve your problem if you have multiple Jenkins instances notifying Stash. Plus, you might want to see a reference to all builds for verification afterwards.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -32,14 +32,18 @@ import hudson.tasks.Publisher;
 import hudson.util.FormValidation;
 import hudson.util.Secret;
 import jenkins.model.Jenkins;
+import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
+
 import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHost;
 import org.apache.http.HttpResponse;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.HttpClient;
+import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
+import org.apache.http.client.methods.HttpRequestBase;
 import org.apache.http.config.Registry;
 import org.apache.http.config.RegistryBuilder;
 import org.apache.http.conn.HttpClientConnectionManager;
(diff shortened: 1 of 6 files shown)


scaytrase:
Is this the same option as 'keep repeated builds' in its basics? Disabling this option will replace all previous builds for the same commit and project key.

laurentverbruggen:
The difference is that disabling 'keep repeated builds' will throw away any history of successful/failed builds since you only keep the last one. This pull request allows you to see the history and still be able to merge it if the last one was successful.

scaytrase:
Can you provide any use-case for keeping build history for single chanset (commit)? Build state is the point of change (for me) for three reasons:
 - Source code changes
 - Build configuration changes
 - Dependency changes

None of theese leads to having store all builds for single changeset.

laurentverbruggen:
Well actually the latter two. Why wouldn't you want to store all builds for those cases? You can use them as reference for problem solving.

scaytrase:
While the build configuration is not a part of the source code (like Ant's `build.xml`), and configurations changes are not stored by themselves and could not be referenced - there is no sense to track build according to configuration changes. You never figure out, which change made build fail or pass. 
If it is part of source code - this will work automatically as new commit would have it's own build history.

Dependency changes tracking is matter of strategy. Usually I prefer to test software against lowest declared dependencies and against hightest one. There is no need to test against the whole matrix of bugfix versions of dependencies. Minor versions are usually enough and matrix project (with versions as labels) solves this problem.

The lower dependency is usually fixed with your project dependency map and it's change is the source code change. So the bumping lower dependency will reset your build history.

The highest (up to date) dependency... Well, if Is has changed (and it is not pinned with your project config), I'm not sure if it worth storing info about previous one. It is now outdated.

For problem solving I would rather prefer have a parametrized build which accept concrete dependency list which leads to project failures. With this concrete list it's now worth tracking code changes until it's not resolved.

laurentverbruggen:
Take the following use-case: you have 2 checks for each pull request (by two different CI jobs) and 1 fails, disabling 'keep repeated builds' will only keep the last which you don't want (because one failed and you don't know which one will be kept). But the failed build was because of a glitch on jenkins. Now this situation would require a new commit without this pull request. With it, you might create a process to handle this without the need for a new commit.

scaytrase:
Nope. Each job has own project `key' (see [stash documentation](https://developer.atlassian.com/stash/docs/latest/how-tos/updating-build-status-for-commits.html)). So re-run single job will erase only this job history (with `keep repeated builds` disabled). Other project keys shoulb be left untouched.

Just re-run failed job with `keep repeated builds` disabled and everything should be ok.

laurentverbruggen:
Hmm, well I guess I'll have to agree with you then. They like to see the history here and I assumed others would like this as well. We are running a SNAPSHOT version of this pull request on your jenkins and they are satisfied.

scaytrase:
@laurentverbruggen  So, for now your PR is only aggregating build history, forcibly marking all the builds for single project successfull (while they are present at the jenkins build history)?

laurentverbruggen:
Yes, due to imo a flaw in Stash...

Issue #57, opened 2015-04-28
UnsupportedSchemeException: http protocol is not supported while notifying Stash


mpkusnierz:
I am getting this exception with v1.8 of the Stash Notifier plugin. But it impacts some jobs and not others, currently unable to work out what is different about the jobs to cause this problem. 
The Repository URL is http://stash-my-server:port/scm/repo/my-project.git
The Repository browser is set to stash, with an equivalent URL: http://stash-my-server:port/projects/repo/repos/my-project/browse
The Notify Stash Instance section has got the Stash base URL set to blank; but this works for other jobs using the same jenkins server and stash instance.

Caught exception while notifying Stash with id {HASH}
org.apache.http.conn.UnsupportedSchemeException: http protocol is not supported
	at org.apache.http.impl.conn.DefaultHttpClientConnectionOperator.connect(DefaultHttpClientConnectionOperator.java:108)
	at org.apache.http.impl.conn.BasicHttpClientConnectionManager.connect(BasicHttpClientConnectionManager.java:338)
	at org.apache.http.impl.execchain.MainClientExec.establishRoute(MainClientExec.java:388)
	at org.apache.http.impl.execchain.MainClientExec.execute(MainClientExec.java:236)
	at org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:184)
	at org.apache.http.impl.execchain.RetryExec.execute(RetryExec.java:88)
	at org.apache.http.impl.execchain.RedirectExec.execute(RedirectExec.java:110)
	at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:184)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:82)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:107)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:55)
	at org.jenkinsci.plugins.stashNotifier.StashNotifier.notifyStash(StashNotifier.java:556)
	at org.jenkinsci.plugins.stashNotifier.StashNotifier.processJenkinsEvent(StashNotifier.java:207)
	at org.jenkinsci.plugins.stashNotifier.StashNotifier.prebuild(StashNotifier.java:160)
	at hudson.model.AbstractBuild$AbstractBuildExecution.preBuild(AbstractBuild.java:816)
[...]

PR #76, opened 2015-11-18
Fix the issue when connect the SSL stash server via an http proxy.

ghawkgu:
This fixes the following connection setting. Which is mentioned in the issue #57 
```
stash notifier plugin <-> http proxy <-> stash server with https
```


scaytrase:
@ghawkgu thank you, I'll try to test this fix a bit later. Could you please retarget your PR to the *develop* branch?

ghawkgu:
@scaytrase Thank you for quick response. I'll create a new PR to change the target branch.

Issue #79, opened 2015-11-30
Cannot notify Stash! (Jenkins Root URL not configured)


marianuspotratz:
I get the message `Cannot notify Stash! (Jenkins Root URL not configured)` with the following versions:
Jenkins: 1.625.2
StashNotifier: 1.9.0

No clue, where this comes from, because the following script executes just fine on my instance's script console:
```
Jenkins.instance.getRootUrl()

Result: https://jenkins/
```

Any ideas?

scaytrase:
Not sure but this could be an invalid url\hostname etc

```
// exit if Jenkins root URL is not configured. Stash build API
		// requires valid link to build in CI system.
		if (Jenkins.getInstance().getRootUrl() == null) {
			logger.println(
					"Cannot notify Stash! (Jenkins Root URL not configured)");
			return true;
		}
```

aaronjwood:
I am running into the same thing. Jenkins 1.625.2 and Stash 4.1.0. I get a valid URL when checking on the script console. The URL is formatted like `http://subdomain.domain.com:8080/`

aaronjwood:
Here are some logs:

Error while serving http://<MYURL>:<MYPORT>/job/<MYPROJECT>/descriptorByName/org.jenkinsci.plugins.StashBranchParameter.StashBranchParameterDefinition/fillRepositoryItems
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.kohsuke.stapler.Function$InstanceFunction.invoke(Function.java:298)
	at org.kohsuke.stapler.Function.bindAndInvoke(Function.java:161)
	at org.kohsuke.stapler.Function.bindAndInvokeAndServeResponse(Function.java:96)
	at org.kohsuke.stapler.MetaClass$1.doDispatch(MetaClass.java:121)
	at org.kohsuke.stapler.NameBasedDispatcher.dispatch(NameBasedDispatcher.java:53)
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:746)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:876)
[...]

PR #82, opened 2015-12-09
Provide fallback for getting the instance's root URL

### LINKED ISSUE #81: Try to implement fallback to JenkinsLocationConfiguration
Despite of the fact it URL was successfully configured at the global jenkins configuration `Jenkins.getInstance.getRootUrl()` returns `null`. `JenkinsLocationConfiguration` is said to return correct value at the same time

Should fix #79 

http://stackoverflow.com/questions/15949946/programatically-access-jenkins-url

aaronjwood:
Should fix #81 (started from #79)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -37,6 +37,7 @@ import hudson.tasks.Publisher;
 import hudson.util.FormValidation;
 import hudson.util.ListBoxModel;
 import jenkins.model.Jenkins;
+import jenkins.model.JenkinsLocationConfiguration;
 import net.sf.json.JSONObject;
 import org.apache.commons.lang.StringEscapeUtils;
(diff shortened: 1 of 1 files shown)


scaytrase:
@aaronjwood Thank you, code looks OK, have you tested that pached plugin resoves your problem? I can build *.hpi for you if you need.

aaronjwood:
You're welcome, I haven't tested this out yet so the .hpi would be nice :)

aaronjwood:
Actually, looks like I will have some time today at work. Don't worry about building the .hpi. I'll do it now

aaronjwood:
Well, so far so good. At the very beginning of my build I'm now getting `Notified Stash for commit with id 64c9310d6f58ce8f7cbb46a47e61a45f3350f1f5`

Will let you know if it notifies properly at the very end.

aaronjwood:
Worked. Got the right notifications in stash on both build time and completion time, and no more errors about the root URL being null :)

scaytrase:
Ok, thank you for testing. I'll merge it to develop and release it later. you can use your snapshot build until then

Issue #86, opened 2015-12-22
Empty Credentials Dropdown


stevengssns:
Hi,

we are moving to an updated Jenkins environment (latest versions), where my team has received a Folder (com.cloudbees.hudson.plugins.folder.Folder) which we can configure as we wish (e.g. adding and configure jobs, credentials, etc..).
Our administrator is using this setup to allow better access control management, which is fine.

However, there seems to be an issue with the StashNotifier, as we are not able to select any of the available credentials, which we are able to configure ourselves on the Folder.
We're certain that the credentials are valid, because we have no problem whatsoever when using the GitSCM plugin. 
Just to be sure I tried just about every possible configuration I could think of (and to which I have access.. I'm not allowed to change the global configuration for instance).
It's most probably some kind of access rights issue, because our administrator could see the credentials that we configured ourselves, while we couldn't.
He could also select one of the credentials, but still they didn't get used by the build.

So I had a quick look at the Git SCM plugin code which lists the credentials:

https://github.com/jenkinsci/git-plugin/blob/master/src/main/java/hudson/plugins/git/UserRemoteConfig.java

Line 77: doFillCredentialsIdItems

It definitely does things differently than the StashNotifier does it:

https://github.com/jenkinsci/stashnotifier-plugin/blob/develop/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java

Line 497: doFillCredentialsIdItems

The Git SCM plugin is manly using the Item and Item.CONFIGURE to check if a user has sufficient permissions, while the Stash Notifier plugin is using ItemGroup and Computer.CONFIGURE.

I don't have enough knowledge about Jenkins Plugin development to see what's going wrong here, but the issue seems to be (a.o) located here.
Any reason for this different implementation, and has anyone else observed this behaviour?

If necessary, our Jenkins administrator will be able to provide more detailed information about the setup after the holidays.

Kind Regards,
Steven

scaytrase:
stash-notifier accepts only plain user/pass credentials, not all types. Check your crenendials have some of them

stevengssns:
I'm using simple "Username with password" credentials, so that shouldn't be a problem.

scaytrase:
thats right, I'll try to investigate.

scaytrase:
This really looks like odd permission checking. 

@pdodds what do you think of this as Credentials implementor?

stevengssns:
[...]

PR #91, opened 2016-01-12
Computer.CONFIGURE doesn't seem to be the premissions that are requir…

stevengssns:
#86

Computer.CONFIGURE doesn't seem to be the premissions that are required (see javadoc: http://javadoc.jenkins-ci.org/hudson/model/Computer.html)

The Item.CONFIGURE seem to be the appropriate ones (see javadoc: http://javadoc.jenkins-ci.org/hudson/model/Item.html). We are checking if the user has permissions to configure the Item where the StashNotifier post build step has been added, right?

This also seem to correspond (roughly) to how the Git SCM plugin does its checks.
         
In any case this seems to fix the issue on my local setup (cloudbees folder + project matrix based security + local jenkins user database).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -19,17 +19,15 @@ import com.cloudbees.plugins.credentials.Credentials;
 import com.cloudbees.plugins.credentials.CredentialsMatchers;
 import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.common.*;
+import com.cloudbees.plugins.credentials.common.UsernamePasswordCredentials;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
-import hudson.EnvVars;
 import hudson.Extension;
 import hudson.Launcher;
 import hudson.ProxyConfiguration;
 import hudson.model.*;
-import hudson.plugins.git.GitBranchTokenMacro;
 import hudson.plugins.git.Revision;
 import hudson.plugins.git.util.BuildData;
 import hudson.security.ACL;
-import hudson.security.AccessControlled;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.BuildStepMonitor;
 import hudson.tasks.Notifier;
@@ -44,8 +42,7 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHost;
 import org.apache.http.HttpResponse;
-import org.apache.http.auth.AuthScope;
-import org.apache.http.auth.UsernamePasswordCredentials;
+import org.apache.http.auth.*;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.config.Registry;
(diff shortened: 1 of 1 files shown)


scaytrase:
@stevengssns  I'm also not sure, whether we really need these tricks with `context` castings. Could you try to make it look like these examples (and check they work for you)?

This one checks context in more simple way:
https://github.com/jenkinsci/rally-plugin/blob/master/src/main/java/com/jenkins/plugins/rally/RallyPlugin.java#L216-L218

This one does not checks the context at all
https://github.com/jenkinsci/ansible-plugin/blob/master/src/main/java/org/jenkinsci/plugins/ansible/AbstractAnsibleBuilderDescriptor.java#L44-L51

stevengssns:
Ok, I'll have a look if I manage to refactor the code a little as you suggest. 
I just kept it simple because I've got no real knowledge about the Jenkins architecture.

stevengssns:
I wasn't sure if I needed to use "@AncestorInPath AbstractProject project" (like the ansible project you referred to), but because the lookup takes an Item, I figured it would do the necessary searching up the tree and used "@AncestorInPath Item project".
The final code is almost the same as for the git plugin, and it seems to work.

For the second issue I basically made sure the build.getProject() is being used when doing the credentials lookup. Because it was being used in 2 places, I created a private helper method for this.

If there are any remarks, I'm sure I'll hear about them.

Fyi, I noticed I my IntelliJ formatting settings differ from the ones used here. In the code I added the indents are with white spaces instead of tabs.

scaytrase:
code style is not a problem. I'll try to get it to review soon

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -346,14 +343,7 @@ private HttpClient getHttpClient(PrintStream logger, AbstractBuild<?, ?> build)
...
 
-		// Determine if we are using the local or global settings
-		String credentialsId = getCredentialsId();
-		if (StringUtils.isBlank(credentialsId)) {
-			credentialsId = descriptor.getCredentialsId();
-		}
-
-		Credentials credentials = CredentialsMatchers.firstOrNull(CredentialsProvider.lookupCredentials(CertificateCredentials.class,
-				Jenkins.getInstance(), ACL.SYSTEM), CredentialsMatchers.withId(credentialsId));
+        CertificateCredentials certificateCredentials = getCredentials(CertificateCredentials.class, build.getProject());
```
scaytrase: Wrapping CertificateCredentials with the same logic is invalid as you share the same Credentials ID for REST requests and for CertificateCredentials. They would not work together

If you want to avoid errors with self-signed certs - there should be separate credential id for it.

scaytrase:
@stevengssns Rest of diff looks good.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
stevengssns: I'll have another look at it after work this evening.

But I don't really see how the credentialId's were different before my refactoring? 
Maybe the difference is to be found in the "strange" assignment of the credentialsId member variable in the username/password credentials lookup block, which I didn't fully understand?

But maybe it was just the late hour..

scaytrase: Yeah, I've noticed you haven't change much now. It's ok then. I'll consider review it later or reimplement when 2.0 comes

scaytrase:
I'll test your branch and merge it soon.

scaytrase:
@stevengssns thank you for you contibution

scaytrase:
I've tagger a prerelease tag, travis should put a usable snapshot HPI there soon

ma77h1a5:
Is there already a new snaphot available for testing?

stevengssns:
If I'm not mistaken 1.9.1-SNAPSHOT should be the one you are looking for

https://github.com/jenkinsci/stashnotifier-plugin/releases/tag/1.9.1-alpha2

ma77h1a5:
OK thanks. I'm going to give it a try.

ma77h1a5:
I can confirm that job specific credentials work with 1.9.1-alpha2! Thanks :D

scaytrase:
nice, I'll try to make it released on weekend

stevengssns:
Excellent!

stevengssns:
Hi @scaytrase, I see the new release was made, but it's not showing up on the Jenkins CI site (or in the plugins that can be installed).

Is some additional action needed for this? Or might it be related to the fact that there is still "-SNAPSHOT" in the version of the code on the release branch/tag ?

scaytrase:
Looks strange. I hoped plugin will apear in Jenkins maven in several hours. but not
http://maven.jenkins-ci.org:8081/content/repositories/releases/org/jenkins-ci/plugins/stashNotifier/

PR #93, opened 2016-01-20
Add unit tests for stashnotifier

cyburs:
This pull request is sample of unit test cases that we plan to contribute in order to raise the overall coverage to 80%. Code coverage is measured using Cobertura This pull request increases code coverage of library from 18% to 37%.

Metric | Before | After
------ | ------ | -----
Coverage % | 18% | 37%
Lines Covered | 52| 104
Total Lines | 1130 | 1130

Vlad Medvedev
DevFactory - Code Quality Team

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/stashNotifier/StashNotifierTest.java
+++ b/src/test/java/org/jenkinsci/plugins/stashNotifier/StashNotifierTest.java
@@ -177,7 +177,8 @@ public class StashNotifierTest
 
         //when
         sn.getHttpClient(logger, build);
-        //then
+
+		//then
         verify(httpClientBuilder).setSSLSocketFactory(any(SSLConnectionSocketFactory.class));
         verify(httpClientBuilder).setConnectionManager(any(HttpClientConnectionManager.class));
     }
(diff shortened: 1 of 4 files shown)


src/test/java/org/jenkinsci/plugins/stashNotifier/StashNotifierTest.java
```diff
@@ -111,4 +127,59 @@ public void test_prebuild_null_revision() {
 		when(build.getActions(BuildData.class)).thenReturn(Collections.singletonList(mock(BuildData.class)));
 		assertTrue(sn.prebuild(build, buildListener));
 	}
+
+    @Test
+    public void test_build_http_client_with_proxy() throws Exception {
+        //given
+        String address = "109.196.210.110";
```
scaytrase: Not sure we should use any public IP even for mock data. Replace with private network IP please

scaytrase:
Currently each test throw exception which, I think should not be expected. You should either fix the travis testing flow config or check the tests.

cyburs:
Hi Pavel, I fixed tests in accordance with your comments. I changed IP address to the local one and eliminated the problem with exceptions messages. Could you check my commit?

scaytrase:
Looks good

PR #100, opened 2016-02-23
Updates for #97

scaytrase:
Fix #97 
Fix #98

PR #101, opened 2016-03-09
Fix tests for #100

mckramer:
Use PowerMockito to mock Hudson, since #hasPermission is final

hudson.model.Node#hasPermission is final, so Mockito.mock is not capable
of completely overriding it.  Use PowerMockito.mock instead.


scaytrase:
Nice, thank you @mckramer

Issue #92, opened 2016-01-20
Pipeline plugin compatibility (Formerly Workflow plugin)


leth:
I'd be great if we could use this plugin from a `Jenkinsfile`!

The plugin repo has [a compatibility guide](https://github.com/jenkinsci/workflow-plugin/blob/master/COMPATIBILITY.md#plugin-developer-guide).

scaytrase:
How does this handle credentials?

leth:
It looks like they reference them by ID, e.g:
```
git url: 'git@git.local.domain:ps/project.git', credentialsId: '67c3072d-b9a7-44fa-a5aa-560ba9c1662f'
```

jglick:
Probably suffices to implement `SimpleBuildStep` and replace `AbstractBuild` with `Run` everywhere, skipping calls to `TokenMacro` if not in fact passed an `AbstractBuild`.

lindt:
is there any known possible workaround?

Something like?
```
sh 'curl https://stash..../foo'
```

scaytrase:
Sure, yes, you could just use official stash\bitbucket server documentation

https://developer.atlassian.com/stash/docs/latest/how-tos/updating-build-status-for-commits.html

headcrabmeat:
Hi,
Does anyone is working on this? If not please assign it to me

headcrabmeat:
BTW @scaytrase, 
whats the target branch should be? master or relase1.0 I see some refactoring on the master but new classes are not currently in use.

scaytrase:
Use default branch (**release/1.x** currently) as target

headcrabmeat:
To all who are interested in. 

I believe that the usage of stash notifier plugin in pipeline should be changed. 
I think that behavior can be same as in the email plugin. 
You can notify stash any time with any status for any repository you want.

So you add a step in the beginning of the build section and insert

``` groovy
step([$class: 'StashNotifier', buildState: 'INPROGRESS', commitSha1: 'bbbb', credentialsId: '', disableInprogressNotification: false, ignoreUnverifiedSSLPeer: false, includeBuildNumberInKey: false, prependParentProjectKey: false, projectKey: 'cccc', stashServerBaseUrl: 'aaaa'])

```

On failure you just catch the error
like

``` groovy
stage 'build'
try {
     sh 'do the build'
} catch (e) {
[...]

PR #108, opened 2016-05-18
Draft version to support workflow

headcrabmeat:
1. Broken backward compatibility to send In-progress status in freestyle project
2. For testing purpose removed some logic in calling TokenMacro.expandAll(...)
3. Working fine in jenkins2.0 + pipleine
4. Tests are not yet updated.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -22,6 +22,7 @@ import com.cloudbees.plugins.credentials.common.*;
 import com.cloudbees.plugins.credentials.common.UsernamePasswordCredentials;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
 import hudson.Extension;
+import hudson.FilePath;
 import hudson.Launcher;
 import hudson.ProxyConfiguration;
 import hudson.model.*;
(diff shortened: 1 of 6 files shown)


ssbarnea:
@headcrabmeat Any updates on this? I am quite interested in reviewing and testing this fix. At this moment it seems that it is not passing CI.

headcrabmeat:
@ssbarnea, 

CI is broken as some tests are not passed but `*.jpi` file could be compiled as well. 
As I wrote above I'd broken some minor features in free-style project kind. 
However you can use this version of the plugin in your Jenkins setup it works well for both pipeline and freestyle.

ariellevy:
any news about this pull request?

PierreBtz:
@headcrabmeat 

I compiled and installed the version with your fix. I can confirm it's working with my freestyle jobs (I did not notice any issue...yet :)). 
I also did a test on a pipeline job using the sample you wrote in #92 and it also worked upon modifying the call to match the name of the enum:

```
public enum StashBuildState {
	SUCCESSFUL, 
	FAILED,
	INPROGRESS,
}
```
I just noticed the plugin does not appear in the pipeline syntax helper. Anyway thanks for your work!

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -604,6 +626,7 @@ public boolean configure(
             prependParentProjectKey = formData.getBoolean("prependParentProjectKey");
 
 			disableInprogressNotification = formData.getBoolean("disableInprogressNotification");
+			buildState = formData.getString("buildState");
```
PierreBtz: As mentioned in #108, this causes a crash in Jenkins global configuration page. I'm not familiar with Jenkins plugin development, but I'm not sure you have to query the buildState here as this method seems to be used only to keep global configurations.

scaytrase: Also I think this is a bad architectural design to pass the build state to the notifier constructor. That's not a part of configuration, but a part of runtime call

scaytrase:
Implemented with #115

PR #111, opened 2016-06-30
Adding a timeout for the url request so the plugin will not hang forever

pHallstedt:
Adding a timeout for the url request so the plugin will not hang forever if the Stash server does not respond

This solves JENKINS-36346 according to my tests...


scaytrase:
[stashNotifier-1.10.4.hpi.zip](https://github.com/jenkinsci/stashnotifier-plugin/files/368429/stashNotifier-1.10.4.hpi.zip)

snapshot build from the jenkins-ci cloudbees CI system

scaytrase:
@pHallstedt please, rebase the branch in order I could test it

PR #116, opened 2016-07-28
Support not built state

aaronjwood:
We don't want to indicate failure for jobs where the NOT BUILT status is encountered. Also, there may be extensions that make use of this result state. This extension is one example that I know of which makes use of the NOT BUILT state https://wiki.jenkins-ci.org/display/JENKINS/Pathignore+Plugin


candita:
Looks good Aaron.

scaytrase:
Looks good, thanks, @aaronjwood

aaronjwood:
Anytime! :)

PR #121, opened 2016-08-16
Update readme.md

sdmoralesma:
added real case example of how to use this plugin with Jenkins' Pipelines

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -40,6 +40,23 @@ current build manually in the Pipeline script.
 
 ```groovy
 node {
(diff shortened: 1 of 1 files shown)


scaytrase:
I hoped to left the initial example alongside the new one as it is actually works for simple setups where no configuration is needed. Could you append it instead if replacing?

sdmoralesma:
updated, let me know if you need other changes.

scaytrase:
Yeah, looks good, thanks @sdmoralesma

PR #126, opened 2016-09-12
Add override fields for failed, succesful, and in-progress messages

pixman20:
Added the ability to override the message rather than using the default of the build description.
The build description can be HTML, which looks really ugly as a message in Stash/BitBucket.

This provides the ability to use variable substitution and set a message for the notification in the config based on the state of the build.


src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -660,6 +688,24 @@ public boolean configure(
 		}
 	}
 
+	@Extension
+	public static class BuildResultTokenMacro extends DataBoundTokenMacro {
```
scaytrase, line 712: @kutsal Macro extension is still present here, but is gone from fields description. Is it still used?

kutsal: It's up to the user to specify it in one of the fields if they like, i.e. they can set the success message to be "Build ${BUILD_NUMBER} on ${NODE_NAME} was ${BUILD_RESULT}". This is particularly useful since a "successful" build may in fact be unstable as tests may have failed, or aborted by user, etc.

I'll add documentation to that effect in the fields' help files and update the PR.

scaytrase:
Sorry for long delay. Generally OK, but for some reason I cannot make this working with pipeline configuration. Can you help me?

Step is configured like following

```groovy
step([$class: 'StashNotifier', successfulMessage: '${BUILD_NUMBER} OK ON ${NODE_NAME}', failedMessage: '${BUILD_NUMBER} NOT OK ON ${NODE_NAME}']) 
```

scaytrase:
I've found the configuration problem, but the newer one leads me to the exception

```
Caught exception while notifying Stash with id e52bee5f6d7cac82e5897964fa69b0d705f99514
java.lang.ClassCastException: org.jenkinsci.plugins.workflow.job.WorkflowRun cannot be cast to hudson.model.AbstractBuild
	at org.jenkinsci.plugins.stashNotifier.StashNotifier.expandMessage(StashNotifier.java:1005)
	at org.jenkinsci.plugins.stashNotifier.StashNotifier.newStashBuildNotificationEntity(StashNotifier.java:885)
	at org.jenkinsci.plugins.stashNotifier.StashNotifier.notifyStash(StashNotifier.java:730)
	at org.jenkinsci.plugins.stashNotifier.StashNotifier.processJenkinsEvent(StashNotifier.java:315)
	at org.jenkinsci.plugins.stashNotifier.StashNotifier.perform(StashNotifier.java:268)
	at org.jenkinsci.plugins.stashNotifier.StashNotifier.perform(StashNotifier.java:242)
	at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:69)
	at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:59)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1$1.call(AbstractSynchronousNonBlockingStepExecution.java:52)
	at hudson.security.ACL.impersonate(ACL.java:221)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1.run(AbstractSynchronousNonBlockingStepExecution.java:49)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

Issue #107, opened 2016-05-16
Override the description


vodik:
It looks like this plugin reads the build's description - if available - and forwards it to stash. However, I'm already using (abusing?) the build description field to have a URL that points back to the triggering PR in Stash. This way I can nicely jump back and forth.

The problem, though, is that means that Stash's build status pane contains escaped html (e.g. `<a href='http://git.example.org/projects/foo/repos/bar/pull-requests/11...`).

It be nice if I could just set a static field to override this (e.g. 'Unit tests for foo...') instead since this link doesn't make much sense.

scaytrase:
I think the option to override the description alongside the project key and other properties would be nice.

kutsal:
I've attempted an implementation of this. Please review PR #127 at your convenience.

scaytrase:
I'm closing this since no activity on proposed PR. Feel free to revive it.

PR #127, opened 2016-09-15
Implementation of notification override mechanism

kutsal:
Implementing issue/request #107:

* Added "useJobDescription" boolean field
* Added "description" TokenMacro-compatible field to override notification sent
* Added TokenMacro macro BUILD_RESULT containing Jenkins build result

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -64,6 +64,7 @@ import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.client.ProxyAuthenticationStrategy;
 import org.apache.http.impl.conn.BasicHttpClientConnectionManager;
 import org.apache.http.util.EntityUtils;
+import org.jenkinsci.plugins.tokenmacro.DataBoundTokenMacro;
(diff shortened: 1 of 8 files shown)


scaytrase:
Also I'm calling @pixman20 here to discuss compatibility with #126 and vice-versa

kutsal:
Yep, I just saw his PR, and I think I like his approach better -- separate message templates for different states. Only difference with mine is the added BUILD_RESULT TokenMacro that gets used as default when nothing is specified anywhere.

pixman20:
I do prefer separate message overrides myself as well, although if I had time it'd be nice to be able to select to add an override via an "Add Override" button, which would contain a Jenkins build status, the override string, and a Stash Build Status (in order to override the mapped status in cases where a state like UNSTABLE or ABORTED in Jenkins should actually be shown as a SUCCESS in Stash, etc.).  
This is more work, but I bring it up because both of our PRs would make that more functional change to be a breaking change, whereas right now it would not be if that was done instead.

All that said, I do really like the ability to use BUILD_RESULT.  After I made my PR I actually started wishing I had that as well :).
My preference would be to use the overrides, but then add in the BUILD_RESULT macro expansion capability.

Thoughts?

kutsal:
@pixman20 I can merge my BUILD_RESULT change into your PR and close this one.

kutsal:
Closing this PR since @pixman20 merged my TokenMacro addition into his PR #126 which is a better implementation of this.

PR #145, opened 2017-02-22
added pipeline compatibility

thebignet:
Plugin token-macro add pipeline compatibility as of version 2.0.  getBuildKey was failing with `Unable to expand build key macro with run of type org.jenkinsci.plugins.workflow.job.WorkflowRun` when notifying in pipeline

PR #146, opened 2017-02-22
add specific description field overriding build description

thebignet:


scaytrase:
@thebignet looks OK but, I'm personally idea, which was tried to be implemented here
https://github.com/jenkinsci/stashnotifier-plugin/pull/126. Could you try to finish that PR instead (I think, integrating it with #145 may help)?

thebignet:
It might be more flexible to be able to set full description (independently of state).  I've added #147 to set state also.

scaytrase:
The problem that it is too hard to manipulate with description text within non-pipeline build, so predefined options are more portable

PR #147, opened 2017-02-22
added state parameter

thebignet:
Possibility to override state upon stash notification

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -82,6 +82,7 @@ def notifyStash(String state) {
           includeBuildNumberInKey: false,
           prependParentProjectKey: false,
           projectKey: '',
+          state: 'SUCCESSFUL',
           stashServerBaseUrl: 'https://stash.company.com'])
 
 }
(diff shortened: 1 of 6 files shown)


scaytrase:
Though I understand the purpose of this change, I personally dislike this as architecturally incorrect idea. 

Is it possibly to do things like the following with pipelines (just java-like sample)?

```java
   StashNotifier notifier = new StashNotifier(/** <...> */);
   notifier.notify(StashBuildState.FAILED, 'The build was failed on ${NODE_NAME}');
```

I understand that this might not be possible because of the nature of the build wrappers, but passing the state as the wrapper constructor argument looks weird

trevjonez:
This can be super useful in the case that a single pipeline script reports back multiple status checks. say for example you do three stages in this order `build`, `test`, `lint`. 

It is perfectly reasonable to expect that the `build` stage and `lint` stage pass but the `test` phase did not. By not allowing a custom parameter and only depending on the state of the overall pipeline status you are left with not being able to correctly report back the status of each stage. 

Without being able to specify the state parameter this plugin falls on its face in anything but the trivial use case.

scaytrase:
I fully agree with the reasons behind this pull-request, but in my mind this approach have architectural drawbacks. Don't we really have any way to pass the state i.e as function argument instead of constructor? Constructor is more for initial configuration design in my mind

trevjonez:
On that note more than just the state should probably be moved out to use `@DataBountSetter`

scaytrase:
Sounds a bit better, could you provide a patch or sample usage?

trevjonez:
This stuff is all pretty new to me but I'll see if I can make the time to take a swing at it.

Xyaren:
This needs a mention in the Readme.

PR #148, opened 2017-02-25
ReadMe: Add setup instructions

refactorsaurusrex:
Added additional setup instructions for users new to Jenkins who may not know where to set the Stash Root Url and Credentials, both of which are necessary for this plugin to work. This is particularly relevant for pipeline users, who do not have access to pre- and post-build steps in the Jenkins GUI. Included a Jenkins config GUI screenshot.


scaytrase:
Looks good, thank you @refactorsaurusrex

PR #155, opened 2017-05-01
Jenkinsfile

scaytrase:
Add Jenkinsfile

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,12 @@
+node('maven') {
+   stage('Preparation') {
+      checkout scm
+   }
+   stage('Build') {
+      sh "mvn -Dmaven.test.failure.ignore clean package"
+   }
+   stage('Results') {
+      junit '**/target/surefire-reports/TEST-*.xml'
+      archive 'target/*.jar'
+   }
+}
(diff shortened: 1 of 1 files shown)


daniel-beck:
@scaytrase Is this meant as a Jenkinsfile for ci.jenkins.io, or for another Jenkins instance?

I'm asking because there's no node with that label on that instance (https://github.com/jenkins-infra/documentation/blob/master/ci.adoc) and we recently opened up Jenkinsfile based CI for all plugins in the jenkinsci org, so now there are builds stuck in the queue.

scaytrase:
actually for my own, but could be replaced. we have a PR for this 

https://github.com/jenkinsci/stashnotifier-plugin/pull/165

If you verify it's correct - it could be merged immediatelly

daniel-beck:
@scaytrase Sorry, didn't see that. Behavior should be similar, so IMO that's good to go.

scaytrase:
merged that PR, tell if something go wrong

daniel-beck:
@scaytrase 

https://ci.jenkins.io/job/Plugins/job/stashnotifier-plugin/ looks good 👍

PR #159, opened 2017-05-10
Ability to ignore or succeed by config settings

lucamilanesio:
Introduce two new configuration settings to allow to control the
Stash notification in case of build aborted or unstable:

- disableAbortAsFailure: do not notify anything if build is aborted
- considerUnstableAsSuccess: when build is finished with some failures,
  consider the stash notification as a success

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -129,6 +129,9 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
 	/** whether to send INPROGRESS notification at the build start */
 	private final boolean disableInprogressNotification;
 
+	/** whether to consider UNSTABLE builds as failures or success */
+	private final boolean considerUnstableAsSuccess;
+
 	private JenkinsLocationConfiguration globalConfig = new JenkinsLocationConfiguration();
 
 // public members ----------------------------------------------------------
@@ -147,7 +150,8 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
 			boolean includeBuildNumberInKey,
 			String projectKey,
 			boolean prependParentProjectKey,
-			boolean disableInprogressNotification
+			boolean disableInprogressNotification,
+			boolean considerUnstableAsSuccess
 	) {
 
 
@@ -162,6 +166,7 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
 		this.projectKey = projectKey;
 		this.prependParentProjectKey = prependParentProjectKey;
 		this.disableInprogressNotification = disableInprogressNotification;
+		this.considerUnstableAsSuccess = considerUnstableAsSuccess;
 	}
 
 	public boolean isDisableInprogressNotification() {
@@ -226,14 +231,20 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
 
 		PrintStream logger = listener.getLogger();
 
-		Result result = run.getResult();
-		if (result == null && disableInProgress) {
+		Result buildResult = run.getResult();
+		if (buildResult == null && disableInProgress) {
 			return true;
-		} else if (result == null) {
+		} else if (buildResult == null) {
 			state = StashBuildState.INPROGRESS;
-		} else if (result.equals(Result.SUCCESS)) {
+		} else if (buildResult == Result.SUCCESS) {
 			state = StashBuildState.SUCCESSFUL;
-		} else if (result.equals(Result.NOT_BUILT)) {
+		} else if (buildResult == Result.UNSTABLE && considerUnstableAsSuccess) {
+			logger.println("UNSTABLE reported to stash as SUCCESSFUL");
+			state = StashBuildState.SUCCESSFUL;
+		} else if (buildResult == Result.ABORTED && disableInProgress) {
+			logger.println("ABORTED");
+			return true;
+		} else if (buildResult.equals(Result.NOT_BUILT)) {
 			logger.println("NOT BUILT");
 			return true;
 		} else {
--- a/src/test/java/org/jenkinsci/plugins/stashNotifier/StashNotifierTest.java
+++ b/src/test/java/org/jenkinsci/plugins/stashNotifier/StashNotifierTest.java
@@ -60,6 +60,7 @@ import java.util.List;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.CoreMatchers.containsString;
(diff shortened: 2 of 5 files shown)


scaytrase:
What about leaving builds blue (inprogress) in case if `disableAbortAsFailure` is set to true and build was aborted? I think that actually this flag should behave equally with `disableInprogressNotification`  since there is no way to 'undo' the build from the bitbucket server now. 

So if the inprogress is off - disable abort failure too. if it is enabled - mark it red upon abort. 

What do you think, @lucamilanesio ?

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -226,14 +236,17 @@ private boolean perform(Run<?, ?> run,
...
-		if (result == null && disableInProgress) {
+		Result buildResult = run.getResult();
+		if (buildResult == null && disableInProgress) {
 			return true;
-		} else if (result == null) {
+		} else if (buildResult == null) {
 			state = StashBuildState.INPROGRESS;
-		} else if (result.equals(Result.SUCCESS)) {
+		} else if (buildResult == Result.SUCCESS ||
+				(buildResult == Result.UNSTABLE && considerUnstableAsSuccess)) {
```
scaytrase: maybe additional `else if` branch? would be easier to read

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -226,14 +236,17 @@ private boolean perform(Run<?, ?> run,
...
 			state = StashBuildState.SUCCESSFUL;
-		} else if (result.equals(Result.NOT_BUILT)) {
+		} else if (buildResult == Result.ABORTED && disableAbortAsFailures) {
+			return true;
```
scaytrase: I think we miss logging here too,

scaytrase (changes requested):
I think we also miss some tests for new behavior

lucamilanesio:
@scaytrase yes, if disableInprogressNotification is true, then aborted build shouldn't do anything. Let me fix that.

lucamilanesio:
@scaytrase I've applied the changes, including the extra tests. Let me know if you need me to add/change anything else. Thanks for your review.

src/test/java/org/jenkinsci/plugins/stashNotifier/StashNotifierTest.java
```diff
@@ -306,6 +322,41 @@ public void test_perform_build_step_success() throws Exception {
...
+        //then
+        ArgumentCaptor<String> messageCaptor = ArgumentCaptor.forClass(String.class);
+        verify(logger, atLeastOnce()).println(messageCaptor.capture());
+        List<String> values = messageCaptor.getAllValues();
+        assertThat(values.get(0), is(containsString("UNSTABLE reported to stash as SUCCESSFUL")));
+        assertThat(values.get(1), is(containsString("Notified Stash for commit with id")));
+    }
+
+    @Test
+    public void test_perform_build_step_abourted_without_notifying_stash() throws Exception {
```
scaytrase: Little typo in `abourted`

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -226,14 +236,20 @@ private boolean perform(Run<?, ?> run,
...
+		} else if (buildResult == null) {
 			state = StashBuildState.INPROGRESS;
-		} else if (result.equals(Result.SUCCESS)) {
+		} else if (buildResult == Result.SUCCESS) {
 			state = StashBuildState.SUCCESSFUL;
-		} else if (result.equals(Result.NOT_BUILT)) {
+		} else if (buildResult == Result.UNSTABLE && considerUnstableAsSuccess) {
+			logger.println("UNSTABLE reported to stash as SUCCESSFUL");
+			state = StashBuildState.SUCCESSFUL;
+		} else if (buildResult == Result.ABORTED && (disableAbortAsFailures || disableInProgress)) {
```
scaytrase: Should this be `(disableAbortAsFailures && disableInProgress)` ? We should not allow the notifier to leave the build blue, so if we have not disable inprogress thus we cannot drop aborted without marking?

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -147,7 +153,9 @@ public StashNotifier(
 			boolean includeBuildNumberInKey,
 			String projectKey,
 			boolean prependParentProjectKey,
-			boolean disableInprogressNotification
+			boolean disableInprogressNotification,
+			boolean disableAbortAsFailures,
```
scaytrase: I think this should be `disableAbortAsFailure` in a single form

scaytrase (changes requested):
Two minor issues and one logic question left. Everything else is OK

src/test/java/org/jenkinsci/plugins/stashNotifier/StashNotifierTest.java
lucamilanesio: Oops, fixing it now :-)

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
lucamilanesio: I guess we should change it in all the other places as well ;-)

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
lucamilanesio: Hold on, if we disable the in progress *and* the build has just being canceled, why should we mark it as failed after all?
On the other side you are right, if we started saying that the build is in progress *and* has been canceled, we have to tell Stash somehow.

Does this mean that we do not need any disableAbortAsFailure *and* we just notify the abort only if we informed Stash that the build was in progress?

scaytrase: > why should we mark it as failed after all

Sounds reasonable. I it is equal to disable inprogress indeed

lucamilanesio: Done, just removed the flag and used the current inprogress configuration.

scaytrase (approved):
Generally looks OK, I'll test and merge is as soon as I can

lucamilanesio:
@scaytrase did you have any time to do a smoke test yourself? We already use it here in production and works :-)

danielcherubini:
I seem to be missing the `considerUnstableAsSuccess` setting in jenkins config.

has something gone wrong?

PR #172, opened 2017-10-03
Added Declarative Pipeline example

pongells:
When using declarative pipelines, a successful build will set `currentBuild.result = null`, which is not what this plugin expects.


scaytrase:
thank you, @pongells

Issue #136, opened 2016-11-23
Pretty DSL alias for pipeline


scaytrase:
See https://issues.jenkins-ci.org/browse/JENKINS-39948

scaytrase:
fixed with #177

PR #177, opened 2017-10-19
enable dsl support

bakito:
This pull request implements a pretty dsl support for the stash notifier plugin as requested in https://github.com/jenkinsci/stashnotifier-plugin/issues/136.

For the usage see readme.md

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -49,7 +49,7 @@ current build manually in the Pipeline script.
 node {
     checkout scm                            // Necessary so we know the current commit
 
-    notifyStash()                           // Notifies the Stash Instance of an INPROGRESS build
+    notifyBitbucket()                       // Notifies the Stash Instance of an INPROGRESS build
 
     try {
         // Do stuff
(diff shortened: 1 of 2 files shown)


scaytrase:
hi, @bakito thank you for contribution!

Not familiar with DSL. Can we use two symbols at once, i.e have two aliases - `notifyStash`, `notifyBitbucket` ? I think this will be much easier to understand for new users to use bitbucket named alias. Other names are still not changed for BC, but new things can reflect actual names

scaytrase (approved):

bakito:
Hi @scaytrase
Although the @Symbol annotation allows to define String[] as value, only the first one is used as DSL identifier.

I could change it to notifyBitbucket.

bakito:
Correction: Both aliases are interpreted, but only the fist one is considered in the Pipeline Syntax wizard.

I could use notifyBitbucket as first and notifyStash as second.

scaytrase (approved):

bakito:
How do poll request work here? Do I have to do someting or are we just waiting for merge?

scaytrase:
@bakito All fine, I'll test your PR and merge it soon. Will be available with next release

scaytrase:
Thank you, @bakito

bakito:
You're welcome.
I guess you can close https://github.com/jenkinsci/stashnotifier-plugin/issues/136 now

PR #181, opened 2017-11-21
Support SHA-1 commit lookup for all runs

ColMelvin:
When given an AbstractBuild Run, the lookupCommitSha1s() function had the
information for TokenMacro to expand the SHA-1 commits.  However, if the
Run was a non-AbstractBuild object, there was not enough information for
TokenMacro to perform expansion (at least, with the parameters originally
provided).

Because of this, warnings were displayed by the StashNotifier step on some
of the runs.  Quite often, StashNotifier is one of the last steps of a
build, so the StashNotifier warning was right next to the final message
about a build failure.  And since the two "errors" were right next to each
other in the log, new users looking at the Console Output often came to
the incorrect conclusion that StashNotifier was the cause of the error.

Fortunately, TokenMacro has a second function to expandAll macros, one
that works for non-AbstractBuilds.  It requires an additional parameter,
but we are given a copy of it in the perform() routine: the workspace.  By
passing this parameter down the stack, we can enable token expansion on
all types of Runs.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -207,7 +207,7 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
 
     @Override
 	public boolean prebuild(AbstractBuild<?, ?> build, BuildListener listener) {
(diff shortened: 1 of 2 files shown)


scaytrase (approved):

scaytrase:
LGTM, I'll test and merge it as soon as I can

scaytrase:
Thank you, @ColMelvin

PR #185, opened 2017-12-20
Job specific ignore unverified ssl

johnjohndoe:

Issue #179, opened 2017-10-26
Global setting "Ignore unverified SSL certificates" without effect


johnjohndoe:
I noticed that the setting `Ignore unverified SSL certificates` in Jenkins' system configuration is ignored.

# Environment
- Jenkins 1.651.3
- Stash Notifier 1.12
- Self-hosted Stash instance (version unclear)
- `Ignore unverified SSL certificates` enabled in system settings
- `Ignore unverified SSL certificates` disabled in job settings

# Failure

The console log shows the following output for the post-build task:

``` java
Notifying Stash at "https://stash.exmaple.com"
Caught exception while notifying Stash with id 5e150ebf30e3c6fed1d90aff80b21c2ee9bc03fa
javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:192)
	at sun.security.ssl.SSLSocketImpl.fatal(SSLSocketImpl.java:1949)
	at sun.security.ssl.Handshaker.fatalSE(Handshaker.java:302)
	at sun.security.ssl.Handshaker.fatalSE(Handshaker.java:296)
	at sun.security.ssl.ClientHandshaker.serverCertificate(ClientHandshaker.java:1509)
	at sun.security.ssl.ClientHandshaker.processMessage(ClientHandshaker.java:216)
	at sun.security.ssl.Handshaker.processLoop(Handshaker.java:979)
	at sun.security.ssl.Handshaker.process_record(Handshaker.java:914)
	at sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:1062)
	at sun.security.ssl.SSLSocketImpl.performInitialHandshake(SSLSocketImpl.java:1375)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1403)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1387)
	at org.apache.http.conn.ssl.SSLConnectionSocketFactory.createLayeredSocket(SSLConnectionSocketFactory.java:394)
	at org.apache.http.conn.ssl.SSLConnectionSocketFactory.connectSocket(SSLConnectionSocketFactory.java:353)
[...]

PR #186, opened 2017-12-20
Point out which credentials are used by the plugin and which are not.

johnjohndoe:
- Lessons learned in #185 and #179.


scaytrase:
I think the best way is not to allow to select crendentials other than can be used with Bitbucker REST API

johnjohndoe:
I rebased this branch onto the latest commit on `release/1.x` for you to merge easily.

PR #149, opened 2017-03-22
Use DisplayURLProvider to support Blue Ocean URL

htanjo:
This PR updates the build link to the new URL compatible with [Blue Ocean](https://jenkins.io/projects/blueocean/). The link will be redirected to Blue Ocean's page if available.

image

- [Display URL API](https://plugins.jenkins.io/display-url-api)
- [Implement a DisplayURLProvider for Blue Ocean - Jenkins JIRA](https://issues.jenkins-ci.org/browse/JENKINS-37878)

PR #190, opened 2018-01-09
Use DisplayURLProvider for computing the build URL

Wilfred:
If the Jenkins instance has Blue Ocean installed, we should generate a
URL for the relevant part of the Blue Ocean API.

See [JENKINS-37878](https://issues.jenkins-ci.org/browse/JENKINS-37878).

Also update the minimum Jenkins version to match display-url-api:
https://github.com/jenkinsci/display-url-api-plugin/blob/display-url-api-0.1/pom.xml#L16

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.502</version>
+        <version>1.625.3</version>
     </parent>
 
     <artifactId>stashNotifier</artifactId>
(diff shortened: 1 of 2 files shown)


scaytrase:
@Wilfred master is a bit outdated branch for now. Main `release/1.x` branch already contains this fix  since #149 

Thank you, anyway

Wilfred:
Woops, I'm still on v1.11.6. Thanks.

Would it be possible to make `master` the latest branch? That would have helped me avoid this.

scaytrase:
`master` is for future development, for which I have eventually no time (and blue ocean with pipelines invalidated some ideas, so I have to rethink everything)

PR #192, opened 2018-02-27
added customBuildState property.

0xorial:
Something that makes this plugin less coupled to 'currentBuild' global variable.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -95,6 +95,9 @@ def notifyBitbucket(String state) {
 }
 ```
 
+Alternatively you can use `customBuildState` property. To use it set `useCustomBuildState` to `true` and `customBuildState` to one of `['SUCCESSFUL', 'FAILED', 'INPROGRESS']`
+
+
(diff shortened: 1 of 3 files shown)


scaytrase:
This was suggested long ago and I disagree with this approach. 

https://github.com/jenkinsci/stashnotifier-plugin/pull/147

This way makes the constructor just a static function to run notification process as an entry point

0xorial:
Thanks for information! In our case we just need to send the specific notification that is based on what we do inside our Jenkinsfile, so own logic of StashNotifier was only confusing. 

What do you think about following approach?

Split the StashNotifier class in 2: one responsible for generating and sending message to bitbucket, and another one for actual build state logic. This way signature and functionality of StashNotifier stays the same, but those who do not need all of its logic can easily opt-out by using the 'lower-level' class.

scaytrase:
This plugin lack good refactoring indeed. If you can suggest some thoughts or work on it, I think it could be discussed, I'm fully open for good improvements

One of the ways I see to achieve it is to move out general notificaton logic to simplified classes and build two separate notifiers ontop of it - one as a classic build wrapper (freeconf post-build step) and one as pipeline\build step

scaytrase:
I think I can open `release 2.0` discussion issue improvements discussion and close this to not confuse anyone

scaytrase:
@0xorial please see #193

Issue #199, opened 2018-06-21
Bitbucket response is a re-direct (301)


svante66:
Hey,
When using the stashnotifier plugin and the Bitbucket Server is using a re-direct response (301) the plugin fails. Is there a special reason that the plugin only accepts 204 (https://github.com/jenkinsci/stashnotifier-plugin/blob/release/1.x/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java#L749) as response?
If not, I guess also at least 301 should be considered ok
Cheers,
// Svante

scaytrase:
Hi, can can you provide both original and redirected URL to compare? 301 should not be considered OK since it does not keep the POST body (it transforms POST to GET). In general it looks like bitbucket base url misconfiguration, i.e http -> https or trailing slash redirects.

204 Code strictly follows the documentation

https://docs.atlassian.com/bitbucket-server/rest/5.11.1/bitbucket-build-rest.html#idm45186530822752

No codes other than 204, 400, 401 are allowed here

PR #200, opened 2018-06-21
added response code 301 as accepted

svante66:


scaytrase:
As I've stated in #199 301 is not a valid bitbucket rest api response status code and it should not be supported

PR #201, opened 2018-08-21
Add setters to the StashNotifier.DescriptorImpl class

alexchiri:
In order to be able to configure this plugin using the [Configuration As Code plugin](https://github.com/jenkinsci/configuration-as-code-plugin), the StashNotifier.DescriptorImpl needs to have setters or public fields. Using YAML to configure Jenkins is pretty nice and this would enable this to work with this plugin as well.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.36</version>
+        <version>3.40</version>
         <relativePath />
     </parent>
     <packaging>hpi</packaging>
@@ -14,14 +14,6 @@
         <jenkins.version>2.60.3</jenkins.version>
         <!-- Java Level to use. Java 7 required when using core >= 1.612 -->
         <java.level>8</java.level>
-        <!-- Jenkins Test Harness version you use to test the plugin. -->
-        <!-- For Jenkins version >= 1.580.1 use JTH 2.x or higher. -->
-        <jenkins-test-harness.version>2.48</jenkins-test-harness.version>
-        <!-- Other properties you may want to use:
-             ~ hpi-plugin.version: The HPI Maven Plugin version used by the plugin..
-             ~ stapler-plugin.version: The Stapler Maven plugin version required by the plugin.
-        -->
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
     <artifactId>stashNotifier</artifactId>
     <version>${revision}${changelist}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.10</version>
+        <version>3.36</version>
         <relativePath />
     </parent>
     <packaging>hpi</packaging>
@@ -11,9 +11,9 @@
         <revision>1.18</revision>
         <changelist>-SNAPSHOT</changelist>
         <!-- Baseline Jenkins version you use to build the plugin. Users must have this version or newer to run. -->
-        <jenkins.version>1.625.3</jenkins.version>
+        <jenkins.version>2.60.3</jenkins.version>
         <!-- Java Level to use. Java 7 required when using core >= 1.612 -->
-        <java.level>7</java.level>
+        <java.level>8</java.level>
         <!-- Jenkins Test Harness version you use to test the plugin. -->
         <!-- For Jenkins version >= 1.580.1 use JTH 2.x or higher. -->
         <jenkins-test-harness.version>2.48</jenkins-test-harness.version>
@@ -142,6 +142,18 @@
             <artifactId>token-macro</artifactId>
             <version>2.0</version>
             <type>jar</type>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>matrix-project</artifactId>
+            <version>1.9</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.jenkins</groupId>
+            <artifactId>configuration-as-code</artifactId>
+            <version>1.8</version>
+            <scope>test</scope>
         </dependency>
     </dependencies>
     <build>
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
         <java.level>7</java.level>
         <!-- Jenkins Test Harness version you use to test the plugin. -->
         <!-- For Jenkins version >= 1.580.1 use JTH 2.x or higher. -->
-        <jenkins-test-harness.version>2.19</jenkins-test-harness.version>
+        <jenkins-test-harness.version>2.48</jenkins-test-harness.version>
         <!-- Other properties you may want to use:
              ~ hpi-plugin.version: The HPI Maven Plugin version used by the plugin..
              ~ stapler-plugin.version: The Stapler Maven plugin version required by the plugin.
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -660,15 +660,6 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
             this.includeBuildNumberInKey = includeBuildNumberInKey;
         }
 
-        public String getProjectKey() {
-            return projectKey;
-        }
-
-        @DataBoundSetter
-        public void setProjectKey(String projectKey) {
-            this.projectKey = StringUtils.trimToNull(projectKey);
-        }
-
         public boolean isPrependParentProjectKey() {
             return prependParentProjectKey;
         }
@@ -676,6 +667,15 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
         @DataBoundSetter
         public void setPrependParentProjectKey(boolean prependParentProjectKey) {
             this.prependParentProjectKey = prependParentProjectKey;
+        }
+
+        public String getProjectKey() {
+            return projectKey;
+        }
+
+        @DataBoundSetter
+        public void setProjectKey(String projectKey) {
+            this.projectKey = StringUtils.trimToNull(projectKey);
         }
 
         public String getStashRootUrl() {
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -666,7 +666,7 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
 
         @DataBoundSetter
         public void setProjectKey(String projectKey) {
-            this.projectKey =  StringUtils.trimToNull(projectKey);
+            this.projectKey = StringUtils.trimToNull(projectKey);
         }
 
         public boolean isPrependParentProjectKey() {
@@ -679,11 +679,7 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
         }
 
         public String getStashRootUrl() {
-            if ((stashRootUrl == null) || (stashRootUrl.trim().isEmpty())) {
-                return null;
-            } else {
-                return stashRootUrl;
-            }
+            return stashRootUrl;
         }
 
         @DataBoundSetter
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -611,33 +611,65 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
             }
         }
 
+        public void setStashRootUrl(String stashRootUrl) {
+			this.stashRootUrl = stashRootUrl;
+		}
+
         public boolean isDisableInprogressNotification() {
             return disableInprogressNotification;
         }
 
+        public void setDisableInprogressNotification(boolean disableInprogressNotification) {
+			this.disableInprogressNotification = disableInprogressNotification;
+		}
+
         public boolean isConsiderUnstableAsSuccess() {
             return considerUnstableAsSuccess;
+        }
+
+        public void setConsiderUnstableAsSuccess(boolean considerUnstableAsSuccess) {
+            this.considerUnstableAsSuccess = considerUnstableAsSuccess;
         }
 
         public String getCredentialsId() {
             return credentialsId;
         }
 
+        public void setCredentialsId(String credentialsId) {
+			this.credentialsId = credentialsId;
+		}
+
         public boolean isIgnoreUnverifiedSsl() {
             return ignoreUnverifiedSsl;
         }
+
+        public void setIgnoreUnverifiedSsl(boolean ignoreUnverifiedSsl) {
+			this.ignoreUnverifiedSsl = ignoreUnverifiedSsl;
+		}
 
         public boolean isIncludeBuildNumberInKey() {
             return includeBuildNumberInKey;
         }
 
+        public void setIncludeBuildNumberInKey(boolean includeBuildNumberInKey) {
+			this.includeBuildNumberInKey = includeBuildNumberInKey;
+		}
+
         public String getProjectKey() {
             return projectKey;
         }
 
+        public void setProjectKey(String projectKey) {
+			this.projectKey = projectKey;
+		}
+
         public boolean isPrependParentProjectKey() {
             return prependParentProjectKey;
         }
+
+        public void setPrependParentProjectKey(boolean prependParentProjectKey) {
+			this.prependParentProjectKey = prependParentProjectKey;
+		}
 
         public FormValidation doCheckCredentialsId(@QueryParameter String value, @AncestorInPath Item project)
                 throws IOException, ServletException {
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -69,10 +69,7 @@ import org.jenkinsci.Symbol;
 import org.jenkinsci.plugins.displayurlapi.DisplayURLProvider;
 import org.jenkinsci.plugins.tokenmacro.MacroEvaluationException;
 import org.jenkinsci.plugins.tokenmacro.TokenMacro;
-import org.kohsuke.stapler.AncestorInPath;
-import org.kohsuke.stapler.DataBoundConstructor;
-import org.kohsuke.stapler.QueryParameter;
-import org.kohsuke.stapler.StaplerRequest;
+import org.kohsuke.stapler.*;
 
 import javax.annotation.Nonnull;
 import javax.net.ssl.SSLContext;
@@ -311,10 +308,6 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
     private String getRootUrl() {
         Jenkins instance = Jenkins.getInstance();
 
-        if (null == instance) {
-            return globalConfig.getUrl();
-        }
-
         return (instance.getRootUrl() != null) ? instance.getRootUrl() : globalConfig.getUrl();
     }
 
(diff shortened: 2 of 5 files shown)


scaytrase:
All or most of this properties could be set while construction. While should we allow dynamically change the settings during lifecycle. Isn't it enough to pass YAML configuration to constructor in some way?

alexchiri:
@scaytrase Unfortunately, at least as far as [I can see in the plugin implementation](https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/plugin/src/main/java/io/jenkins/plugins/casc/BaseConfigurator.java#L50), it works by discovering fields and setters to determine what properties can be set from the YAML. They have some guidelines for plugin developers [here](https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/PLUGINS.md).  There can be some custom implementation be done on the configuration-as-code plugin side to handle different ways of configuring plugins, but I felt since stash notifier has quite a basic config, that would be a bit of an overkill.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -611,34 +611,66 @@ public String getStashRootUrl() {
             }
         }
 
+        public void setStashRootUrl(String stashRootUrl) {
+			this.stashRootUrl = stashRootUrl;
```
darxriggs: Please use correct indentation for all new code.
Here and in other places spaces & tabs are currently mixed up.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -611,34 +611,66 @@ public String getStashRootUrl() {
             }
         }
 
+        public void setStashRootUrl(String stashRootUrl) {
```
darxriggs: Please annotate all setters with `@DataboundSetter`.
See **Step 1** in https://github.com/jenkinsci/configuration-as-code-plugin/blob/configuration-as-code-1.8/docs/PLUGINS.md.

darxriggs (commented):
Please also use data-binding via `request.bindJson(this, json)` inside `configure(...)`.
See **Step 5** in https://github.com/jenkinsci/configuration-as-code-plugin/blob/configuration-as-code-1.8/docs/PLUGINS.md.

There should also be two tests - one for configuring Jenkins via YAML and one for exporting a configured Jenkins to YAML.
See **How to test?** in https://github.com/jenkinsci/configuration-as-code-plugin/blob/configuration-as-code-1.8/docs/PLUGINS.md.

Note: I am not a maintainer of this plugin. But I already have migrated some other plugins to be compatible with JCasC.

alexchiri:
@darxriggs Thank you for the feedback and review. Will work further on the PR soon.

scaytrase:
@alexchiri as soon I have no views on architectural future of this plugin version (only some thoughts on 2.x version), I think we can accept this PR after suggested impovements

pom.xml
```diff
@@ -3,20 +3,20 @@
...
+        <version>3.36</version>
         <relativePath />
     </parent>
     <packaging>hpi</packaging>
     <properties>
         <revision>1.18</revision>
         <changelist>-SNAPSHOT</changelist>
         <!-- Baseline Jenkins version you use to build the plugin. Users must have this version or newer to run. -->
-        <jenkins.version>1.625.3</jenkins.version>
+        <jenkins.version>2.60.3</jenkins.version>
```
darxriggs, line 14: I suggest  to reduce it to the following snippet:
```
<properties>
    <revision>1.18</revision>
    <changelist>-SNAPSHOT</changelist>
    <jenkins.version>2.60.3</jenkins.version>
    <java.level>8</java.level>
</properties>
```
The rest is already covered by the parent pom nowadays.
(This can be verified with `mvn help:effective-pom`.)

pom.xml
```diff
@@ -3,20 +3,20 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.10</version>
+        <version>3.36</version>
```
darxriggs: If you like, there is already `3.40` - see https://github.com/jenkinsci/plugin-pom/releases.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -637,34 +626,73 @@ public String getStashRootUrl() {
...
+        @DataBoundSetter
+        public void setIncludeBuildNumberInKey(boolean includeBuildNumberInKey) {
+            this.includeBuildNumberInKey = includeBuildNumberInKey;
+        }
+
         public String getProjectKey() {
             return projectKey;
         }
 
+        public void setProjectKey(String projectKey) {
```
darxriggs, line 677: It seems you missed a `@DataBoundSetter` here.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -637,34 +626,73 @@ public String getStashRootUrl() {
             }
         }
 
+        @DataBoundSetter
+        public void setStashRootUrl(String stashRootUrl) {
+            this.stashRootUrl = stashRootUrl;
```
darxriggs: For Strings like this one I suggest to use something like Apache Commons Lang [StringUtils::trimToNull](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#trimToNull-java.lang.String-) or Jenkins [Util::fixEmptyAndTrim](https://javadoc.jenkins.io/hudson/Util.html#fixEmptyAndTrim-java.lang.String-). This improves user experience by better handling strings with leading/trailing spaces.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -637,34 +626,73 @@ public String getStashRootUrl() {
...
+        public void setIncludeBuildNumberInKey(boolean includeBuildNumberInKey) {
+            this.includeBuildNumberInKey = includeBuildNumberInKey;
+        }
+
         public String getProjectKey() {
             return projectKey;
         }
 
+        public void setProjectKey(String projectKey) {
+            this.projectKey = projectKey;
```
darxriggs: See comment about `setStashRootUrl()` on trimming.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -637,34 +626,73 @@ public String getStashRootUrl() {
...
+            this.considerUnstableAsSuccess = considerUnstableAsSuccess;
+        }
+
         public String getCredentialsId() {
             return credentialsId;
         }
 
+        @DataBoundSetter
+        public void setCredentialsId(String credentialsId) {
+            this.credentialsId = credentialsId;
```
darxriggs: See comment about `setStashRootUrl()` on trimming.

src/test/java/org/jenkinsci/plugins/stashNotifier/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,60 @@
...
+        StashNotifier.DescriptorImpl stashNotifierConfig = rule.jenkins.getDescriptorByType(StashNotifier.DescriptorImpl.class);
+
+        assertThat(stashNotifierConfig.getCredentialsId(), equalTo("bitbucket-credentials"));
+        assertThat(stashNotifierConfig.isIgnoreUnverifiedSsl(), equalTo(true));
+        assertThat(stashNotifierConfig.isIncludeBuildNumberInKey(), equalTo(true));
+        assertThat(stashNotifierConfig.getProjectKey(), equalTo("JEN"));
+        assertThat(stashNotifierConfig.isPrependParentProjectKey(), equalTo(true));
+        assertThat(stashNotifierConfig.isDisableInprogressNotification(), equalTo(true));
+        assertThat(stashNotifierConfig.isConsiderUnstableAsSuccess(), equalTo(true));
+    }
```
darxriggs, line 35: Verifying `stashRootUrl` is missing here.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -720,26 +748,20 @@ public boolean configure(
                 StaplerRequest req,
                 JSONObject formData) throws FormException {
 
-            // to persist global configuration information,
-            // set that to properties and call save().
-            stashRootUrl = formData.getString("stashRootUrl");
-            ignoreUnverifiedSsl = formData.getBoolean("ignoreUnverifiedSsl");
-            includeBuildNumberInKey = formData.getBoolean("includeBuildNumberInKey");
+            this.stashRootUrl = null;
```
darxriggs: I suggest to keep the order consistent for easier reading/comparison:
* field declarations in `StashNotifier` (maybe)
* setters/getters in `StashNotifier` (maybe)
* default values (lines 751-758) in `StashNotifier::configure`
* assertThat calls (lines 27-33) in `ConfigAsCodeTest::should_support_jcasc_from_yaml`
* setter calls (lines 40-47) in `ConfigAsCodeTest::should_support_jcasc_to_yaml`

src/test/resources/org/jenkinsci/plugins/stashNotifier/ConfigAsCodeTest/configuration-as-code.yml
```diff
@@ -0,0 +1,10 @@
+unclassified:
+  notifyBitbucket:
+    considerUnstableAsSuccess: true
+    credentialsId: "bitbucket-credentials"
+    disableInprogressNotification: true
+    ignoreUnverifiedSsl: true
+    includeBuildNumberInKey: true
+    prependParentProjectKey: true
+    projectKey: "JEN"
+    stashRootUrl: "https://bitbucket.com"
```
darxriggs: Even if it's only an example value, I suggest to use something like `https://my.company.intranet/bitbucket`. This more clearly states that this plugin is about "Bitbucket Server" and not "Bitbucket Cloud". Also see #225 for a cleanup regarding this.

src/test/java/org/jenkinsci/plugins/stashNotifier/DescriptorImplTest.java
```diff
@@ -73,21 +73,13 @@ public void setUp() throws Exception {
...
-        assertThat(desc.getCredentialsId(), is("someCredentialsId"));
-        assertThat(desc.getProjectKey(), is("someProjectKey"));
-        assertThat(desc.getDisplayName(), is("Notify Stash Instance"));
-        assertThat(desc.isDisableInprogressNotification(), is(true));
-        assertThat(desc.isConsiderUnstableAsSuccess(), is(true));
-        assertThat(desc.isIgnoreUnverifiedSsl(), is(true));
-        assertThat(desc.isIncludeBuildNumberInKey(), is(true));
-        assertThat(desc.isPrependParentProjectKey(), is(true));
-        assertThat(desc.isApplicable(AbstractProject.class), is(true));
+        verify(request).bindJSON(desc, json);
```
darxriggs: The question is, if just verifying a `bindJSON` call is sufficient for you.
To be really sure, values should still be verified (using a `StaplerRequest` implementation/mock that implements `bindJSON` if possible). Otherwise if you for example missed a setter for a field, this would not be caught here and JCasC and UI based configuration would be broken.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -311,10 +308,6 @@ private boolean perform(Run<?, ?> run,
     private String getRootUrl() {
         Jenkins instance = Jenkins.getInstance();
 
-        if (null == instance) {
```
darxriggs, line 314: I think this change is ok.
`Jenkins::getInstance` is now annotated with `Nonnull` and this code only runs when Jenkins is completely initialized.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -531,10 +524,6 @@ public boolean isTrusted(X509Certificate[] chain, String authType)
 
     private void configureProxy(HttpClientBuilder builder, URL url) {
         Jenkins jenkins = Jenkins.getInstance();
-        if (jenkins == null) {
```
darxriggs, line 534: I think this change is ok.
See comment on `getRootUrl()`.

darxriggs (changes requested):
Thanks for the quick update.
I had another look at the new changes. It already looks pretty good.
See my comments for some required changes.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
alexchiri: Yup, and also spotbugs doesn't give you much choice on this. Without this change and the one below, `mvn install` wouldn't complete successfully.

alexchiri:
Thanks again for the detailed feedback and advice on this @darxriggs. Learned some things and feel I'm getting better at it. 😊 

Please let me know if anything else sticks out!

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -629,42 +615,82 @@ public ListBoxModel doFillCredentialsIdItems(@AncestorInPath Item project) {
...
         public boolean isPrependParentProjectKey() {
             return prependParentProjectKey;
         }
 
+        @DataBoundSetter
+        public void setPrependParentProjectKey(boolean prependParentProjectKey) {
+            this.prependParentProjectKey = prependParentProjectKey;
+        }
+
+        public String getStashRootUrl() {
```
darxriggs, line 681: Now that `trimToNull()` is used in `setStashRootUrl()` this can simply return the `stashRootUrl` value (like all other getters).

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -629,42 +615,82 @@ public ListBoxModel doFillCredentialsIdItems(@AncestorInPath Item project) {
...
             return includeBuildNumberInKey;
         }
 
+        @DataBoundSetter
+        public void setIncludeBuildNumberInKey(boolean includeBuildNumberInKey) {
+            this.includeBuildNumberInKey = includeBuildNumberInKey;
+        }
+
         public String getProjectKey() {
             return projectKey;
```
darxriggs: You moved the field `projectKey` before the field `stashRootUrl` but the setter/getter order doesn't match this. But it doesn't really matter.

src/test/java/org/jenkinsci/plugins/stashNotifier/DescriptorImplTest.java
```diff
@@ -74,20 +77,30 @@ public void testConfigure() throws Descriptor.FormException {
         //given
         doNothing().when(desc).save();
 
+        ServletContext servletContext = PowerMockito.mock(ServletContext.class);
```
darxriggs: If possible try not to use PowerMockito but only Mockito. You can have a look in the internet why to avoid PowerMock.

PowerMockito is mainly required to mock constructors and static methods. That's often the case for legacy code that was written without testability in mind. At first sight that's not the case here.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -584,14 +570,14 @@ public DescriptorImpl getDescriptor() {
          * If you don't want fields to be persisted, use <tt>transient</tt>.
          */
 
+        private boolean considerUnstableAsSuccess;
         private String credentialsId;
-        private String stashRootUrl;
+        private boolean disableInprogressNotification;
         private boolean ignoreUnverifiedSsl;
         private boolean includeBuildNumberInKey;
-        private String projectKey;
```
darxriggs, line 591: @scaytrase I am wondring why `projectKey` is a field in `StashNotifier.DescriptorImpl`. It is not visible in the UI and was never part of `global.jelly`. Shouldn't it be removed from `StashNotifier.DescriptorImpl` and only be a field in `StashNotifier`?

darxriggs:
Apart from the `projectKey` field issue, which is to be decided by @scaytrase, this pull request now looks good to me.
I suggest to squash all the commits into one before merging.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
scaytrase: Not sure for it really, may be a copy-paste coincident. We can try to remove it, cannot image when project key should be declared on global level

scaytrase (approved):

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
scaytrase: if it's non blocking for this PR we can try it as a separate change

scaytrase:
thank you @alexchiri ! (and @darxriggs for comments, for sure)

pom.xml
```diff
@@ -143,6 +135,18 @@
             <version>2.0</version>
             <type>jar</type>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>matrix-project</artifactId>
```
proski, line 140: Why was the matrix project dependency added? The commit description makes no mention of it. I don't see the matrix plugin being used in the code. I see that it's only for testing, but I still don't see those new tests.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -21,10 +21,7 @@
 import com.cloudbees.plugins.credentials.common.*;
 import com.cloudbees.plugins.credentials.common.UsernamePasswordCredentials;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
-import hudson.Extension;
-import hudson.FilePath;
-import hudson.Launcher;
-import hudson.ProxyConfiguration;
+import hudson.*;
```
proski, line 24: I don't think wildcards make the code better. There are plenty of tools to keep an up-to-date list of used imports.

proski (commented):
It was surprising to find this large commit in the commit history with a description that makes it sound like a minor change.

pom.xml
darxriggs: That's a `test` scope dependency required by the new tests.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
darxriggs: For me, using wildcard imports or not is a project decision. There are as always advantages and disadvantages. And nowadays tools should be able to handle both.

pom.xml
alexchiri: Sorry, I missed to add that. I added the Matrix project test dependency because the JenkinsRule requires it and it has been removed from core since 1.561. I added the JenkinsRule in the test that verifies the configuration as code setup, according to JCasC documentation. (import and export yaml). (this is the issue where I found more information about the NoClassDefFoundError I got while using the JenkinsRule: https://issues.jenkins-ci.org/browse/JENKINS-27826)

proski: Thank you for the explanation!

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
proski: I agree on all counts, I just wanted to make sure it was an intentional change and not somebody's sloppiness that everybody missed.

scaytrase: I'm not in general a Java developer and I adopted (ocassionaly, huh) this plugin with worse code conventions than it have today, but there is still no CSC strictly defined here today. We can adopt one, I think, if it make life easier.

darxriggs:
@scaytrase I am looking forward for this to be released. :-)

darxriggs:
Just as reference, there is also [JENKINS-56899](https://issues.jenkins-ci.org/browse/JENKINS-56899) tracking this.

PR #213, opened 2018-12-13
Incrementalified

jakub-bochenski:
as per https://github.com/jenkinsci/incrementals-tools/blob/master/README.md#enabling-incrementals-the-easy-way


jakub-bochenski:
I tried to fix the dependency versions to make the new enforcer rule pass, but they are too messy.

I've disabled the check.

jakub-bochenski:
And you can get the version from here https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/stashNotifier/1.15-rc250.b7c39e0ef27f/

scaytrase:
👍 Thank you!

Issue #217, opened 2019-02-05
Support gitscm v3


oakad:
After upgrading (by accident) to GitSCM-v3.0.0-rc we are seeing the following error in our pipeline using `step([$class: 'StashNotifier'])`:

> [Pipeline] notifyBitbucket
> found no commit info

I suppose, this is due to the changes in the new GitSCM, which should be supported.

Issue #27, opened 2014-02-11
Autocomplete shouldn't populate stash notifier config?


pmv:
We put the userid and password for stash notifier in the global config.  Therefore there is no userid or password in the job configuration.  However, if you edit any Jenkins job, in certain situations, a userid and password will automatically be inserted in the job config.

Steps to reproduce:
1) Use either Chrome or Firefox (didn't test IE), and have the browser save your password on the Jenkins 'login' page
2) Edit any job that has stash notifier and save.  You can edit anything - it doesn't have to be related to stash notifier.
3) Look at the config.xml and see your userid and password in the stash notifier section.

Since the userid and password are hidden under an 'Advanced' button, if you don't have it expanded, you won't even notice this has happened when configuring the page.

One obvious fix is to not have your browser store your credentials, but how do you keep all users from doing that?

Would this work?
http://stackoverflow.com/questions/2530/how-do-you-disable-browser-autocomplete-on-web-form-field-input-tag

scaytrase:
I think the problem will gone with #62 soon

PR #218, opened 2019-02-05
Support gitscm v3

klimas7:
Resolving issue #217

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,7 +104,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
-            <version>2.0</version>
+            <version>4.0.0-rc</version>
             <type>jar</type>
         </dependency>
         <dependency>
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -27,7 +27,7 @@ import hudson.Launcher;
 import hudson.ProxyConfiguration;
 import hudson.model.*;
 import hudson.plugins.git.Revision;
-import hudson.plugins.git.util.BuildData;
+import hudson.plugins.git.util.BuildDetails;
 import hudson.security.ACL;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.BuildStepMonitor;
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -104,7 +104,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
-            <version>2.0</version>
+            <version>4.0.0-rc</version>
```
scaytrase, line 107: should we really us 4-RC (not stable) when original issue says about  v3?

klimas7: Hi,
In issue we have a little mismatch. In the same time was release two plugins:
[git-client](https://plugins.jenkins.io/git-client) 3.0.0-rc version
[git](https://plugins.jenkins.io/git) 4.0.0-rc version

Original issue say about git-client version but this plugin use git plugin.
Unfortunately RC versions was deployed to public Jenkins repository and any client can download that versions. I think if any plugins are deployed to public repository is a stable ;)

scaytrase: The latest RC for both were released a week ago. I'm not sure it's a good idea to require this as a dependency at the moment.

Also wouldn't it force users to upgrade their plugins to this RC ones?

klimas7: You're absolutely right, but all user have git plugin and git-client in their Jenkins. Many of them executed jenkins update automatically and got a new versions git and git-client. In that situation stashnotifier plugin doesn't work.

scaytrase: Sounds legit

scaytrase:
I will build new release as soon as I can

scaytrase:
@klimas7 people got some problems with 4-rc

https://issues.jenkins-ci.org/browse/JENKINS-55939

scaytrase:
Can you verify that https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fstashnotifier-plugin/detail/PR-218/1/artifacts HPI works with 4-rc ?

klimas7:
I confirm,  work fine. I checked on this build [#27](https://ci.jenkins.io/job/Plugins/job/stashnotifier-plugin/job/release%252F1.x/27/)

scaytrase:
@klimas7  https://github.com/jenkinsci/stashnotifier-plugin/releases/tag/1.16

scaytrase:
So, we started getting problems, so If we have new features before versions will be stabilized, I'll have to revert this one

https://github.com/jenkinsci/stashnotifier-plugin/issues/221

klimas7:
:( https://github.com/jenkins-infra/update-center2/pull/258 for this situation (I didn't expect that) the best solution will be revert my changes.

scaytrase:
Yeah, looks like we gonna revert

PR #223, opened 2019-02-15
Add note about requiring Jenkins Git plugin

reftel:
Document that only versions 4.0.0-rc and above of the Jenkins Git plugin are supported


scaytrase:
Hi @reftel we are goind to revert 4.0.0-rc requirement, see #222. I'm going to release it on weekend

PR #226, opened 2019-04-01
Remove option to set project key globally (was broken)

darxriggs:
The project key feature was introduced in commit 04a09cc45370.

The implementation had two flaws:
1) It contained `projectKey` in the Descriptor but did not provide this field in the UI (global.jelly).
2) The `getBuildKey()` method only used the global project key in one place. When expanding macros, only the job specific key was used.


scaytrase (approved):

scaytrase:
thank you, @darxriggs !

PR #227, opened 2019-04-02
Improvements

darxriggs:
I suggest to review this commit by commit as each one covers a different aspect.


pom.xml
```diff
@@ -21,7 +21,7 @@
         Jenkins plugin to notify Atlassian Bitbucket Server (formerly known as Stash)
         instances of build results through the REST build API.
     </description>
-    <url>https://wiki.jenkins-ci.org/display/JENKINS/Bitbucket+(Stash)+Notifier+Plugin</url>
+    <url>https://wiki.jenkins.io/display/JENKINS/Bitbucket+(Stash)+Notifier+Plugin</url>
```
scaytrase, line 24: I hope this one won't break the distribution again, as soon as path part is the same. renaming article in wiki broke it long time ago :)

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -387,7 +387,7 @@ private boolean processJenkinsEvent(
         }
 
         // Use a set to remove duplicates
-        Collection<String> sha1s = new HashSet<String>();
```
scaytrase, line 398: Is this possible after bumping java reqs to 8?

src/main/java/org/jenkinsci/plugins/stashNotifier/StashCredentialMatcher.java
```diff
@@ -4,12 +4,13 @@
 import com.cloudbees.plugins.credentials.CredentialsMatcher;
 import com.cloudbees.plugins.credentials.common.CertificateCredentials;
 import com.cloudbees.plugins.credentials.common.UsernamePasswordCredentials;
+import edu.umd.cs.findbugs.annotations.NonNull;
```
scaytrase, line 7: Is it correct import for `NonNull`? Maybe `import javax.annotation.Nonnull;` ?

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -106,7 +106,7 @@
     /**
      * The id of the credentials to use.
      */
-    private String credentialsId;
+    private final String credentialsId;
```
scaytrase, line 109: shouldn't this break recent setters improvement? with setters the whole configuration should be considered mutable

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
darxriggs: That's Java 7 already - https://docs.oracle.com/javase/8/docs/technotes/guides/language/type-inference-generic-instance-creation.html

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
darxriggs: That's the field in `StashNotifier`. There is no setter for it, only a constructor.
The JCasC setters were about the fields in `StashNotifier.DescriptorImpl`.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashCredentialMatcher.java
darxriggs: Better not use `javax.annotation.Nonnull` as there is no official JSR release for it. This is always source of discussions - e.g. see this comment https://github.com/jenkinsci/display-url-api-plugin/pull/19#discussion_r233771043. As written in the commit message I am using exactly the same annotation as the implemented interface is using.

scaytrase (approved):

scaytrase:
Thank you @darxriggs ! It's nice to see fresh action for this plugin. If you have more ideas or improvements - I'm open to them

scaytrase:
I hope I'll release all these changes on this weekend

darxriggs:
@scaytrase will you find some time to release this?

scaytrase:
sure. sorry for delay. got busy week. i'll do it right now

scaytrase:
@darxriggs https://github.com/jenkinsci/stashnotifier-plugin/releases/tag/1.18

darxriggs:
@scaytrase thanks a lot.

PR #237, opened 2019-07-25
pom.xml: Make guava a test requirement

proski:
It is not needed for running the plugin and should not be packaged into
stashNotifier.hpi.

Guava was the largest contributor to the plugin size.

The plugin has been tested on Jenkins to make sure it works and posts notifications to Stash.


scaytrase:
Sounds great improvement

scaytrase:
Thank you, @proski !

PR #238, opened 2019-07-26
Set ConnectionTimeout

### LINKED ISSUE #230: Hangs at "Notifying Stash at ..."
In our environment we're sometimes having some network issues with our Bitbucket Server instance. This seems to cause "Notify Stash Instance" post-build actions to hang. My guess is that there is a missing timeout.
The result is that the build hangs for ever and it is not even possible to cancel from the Jenkins UI.

What we see in the build job console log is:
Notifying Stash at "https://bitbucket.foo.bar"
and then a spinning wheel...

Environment:
Jenkins LTS 2.150.3 on RHEL 7 and openJDK 1.8.0
Stash Notifier plugin v1.17

jinnthegeek:
fixes #230, set connectionTimeout to requestConfig same with socketTimeout


scaytrase (approved):

scaytrase:
We can give it a try, thank you @jinnthegeek !

PR #240, opened 2019-08-01
pom.xml: Fix bundling of sshd-core.jar, update git dependency

proski:
git 2.0 has problematic dependencies that makes the hpi plugin bundle `sshd-code.jar` into the plugin.

Update git dependency to 3.0.0. That fixes the bundling issue.

Thanks to @jglick for suggesting the fix in https://issues.jenkins-ci.org/browse/JENKINS-53957

---

The size of Stash Notifier plugin was totally unreasonable before of bundling. That issue is fully resolved now.

The size of stashNotifier.hpi is 27349 bytes! Yes, it's functional, I've checked it on my test server.

git plugin 3.0.0 was released in September 2016, so this PR doesn't force users to upgrade to anything unreasonably new. I was contemplating using git 3.11.0, but that could make some overly conservative Jenkins admins unhappy.

I don't think the dependency on the git plugin is really needed, but if we remove it, it would be a much bigger change, as its dependencies would need to be listed instead. It can be done separately.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,7 +90,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
-            <version>2.0</version>
+            <version>3.0.0</version>
             <type>jar</type>
         </dependency>
         <dependency>


scaytrase (approved):

scaytrase:
I've inspected linked issue, sounds valid. Thank you @proski

scaytrase:
Unfortunately travis constantly fail on this, I'll try to dig it out. Jenkins builds are ok

proski:
Thank you for merging this PR.

It turns out this plugin depends on Git plugin. `hudson.plugins.git.util.BuildData` is part of Git plugin. So I cannot remove that dependency.

scaytrase:
Yeah, I'm not suggesting removing dependencies, I'm just interested, what's the difference in build process

proski:
OK, I think I understand it. The dependency chain (without test dependencies) was:
```
stashNotifier -> git 2.0 -> git-client 1.4.4 -> ssh-agent 1.3 -> sshd-core 0.8.0
```
Now it is:
```
stashNotifier 1.19 -> git 3.0.0 -> git-client 2.0.0
```
git-client 2.0.0 doesn't depend on ssh-agent, so that cuts the non-test dependency chain.

The version of sshd-core (0.14.0) was determined by a shorter chain that involved test dependencies:
```
stashNotifier 1.19 => plugin 3.40 -> jenkins-war 2.60.3 ~> sshd 1.11 -> sshd-core 0.14.0
```
I'm using `=>` for a child-parent relationship (which doesn't count for the dependency length) and `~>` for a test dependency.

`mvn dependency:tree` was showing a shorter dependency chain, but with the scope based on analyzing the whole tree, which made it look like a test dependency was pulling in a compile dependency.

I don't see a way to make maven omit test dependencies. It would have helped with debugging the issue.

Issue #241, opened 2019-08-06
Add argument for build name


westarne:
An argument to optionally override the build name send in the status notification would be great to make it possible to send one status per build step. This should result in a bitbucket display like this:

![image](https://user-images.githubusercontent.com/15111918/62541095-5d64d480-b859-11e9-86c7-92796592d249.png)

Example usage:
`notifyBitbucket(buildName: "Build")`
Without this parameter I would expect the build name to be derived like before.

scaytrase:
I'm not really sure how it should behave in terms of generating links, etc.

westarne:
I'd expect the plugin to link like before. I'm not sure how the links are generated right now, but all of those entries should generally link to the same build. If possible it could link to the exact build step (which is possible as far as i know for the Blue Ocean Pipeline UI).

scaytrase:
If you don't care about links you can do this kind of setup by yourself by manipulating project key and description while performing notification. different project key will result in different build records

westarne:
Thank you for the advice. I think this will work as a workaround. But in the long term i would look forward to a direct parameter without the need to change build parameters back and forth.

Edit: Actually only the `displayName` can be overwritten. So it is not possible to fully customize the build name this way.

westarne:
Also it is not possible to overwrite `currentBuild.result` multiple times:

```groovy
        currentBuild.result = 'SUCCESS'
        echo currentBuild.result
        currentBuild.result = 'NOT_BUILD'
        echo currentBuild.result
        currentBuild.result = 'SUCCESS'
        echo currentBuild.result
        currentBuild.result = 'NOT_BUILD'
        echo currentBuild.result
        currentBuild.result = 'FAILURE'
        echo currentBuild.result
        currentBuild.result = 'UNSTABLE'
        echo currentBuild.result
        currentBuild.result = 'FAILURE'
        echo currentBuild.result
```

will produce:

```
[Pipeline] echo
SUCCESS
[Pipeline] echo
FAILURE
[Pipeline] echo
FAILURE
[Pipeline] echo
FAILURE
[Pipeline] echo
FAILURE
[Pipeline] echo
FAILURE
[Pipeline] echo
FAILURE
```

This is part of the https://github.com/jenkinsci/stashnotifier-plugin/issues/209 issue.

scaytrase:
@geralvin @westarne I would appreciate a PR for (buildName+buildKey) and\or buildState management with a BC keep. Why not if it solves user problems

[...]

PR #243, opened 2019-08-08
Make build state and build name overwritable

### LINKED ISSUE #209: Add argument for buildResult
I would like to have an argument for the buildResult which is reportet to bitbucket.

My use Case:
From pipeline A I'm stating an integration pipeline B with build(... wait: false, ...). Just after that call pipeline A terminates succesfully. From this moment on, it looks like (in Bitbucket) all the pipelines have been successfully completed. The notification about Pipeline B takes several minutes, until the checkouts in Pipeline B.
My preferred solution: Before calling build() in Pipeline A,  I want to inform Bitbucket about Pipeline B, which is INPROGRESS.

westarne:
Fixes #209 #241

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -87,14 +87,14 @@ Or you could as well use
 ```groovy
     checkout scm
     
-    notifyBitbucket(buildState: 'INPROGRESS')        // Notifies the Bitbucket instance of an INPROGRESS build
+    notifyBitbucket(buildStatus: 'INPROGRESS')        // Notifies the Bitbucket instance of an INPROGRESS build
     
     try {
         // Do stuff
-        notifyBitbucket(buildState: 'SUCCESSFUL')    // Notifies the Bitbucket instance of an SUCCESSFUL build
+        notifyBitbucket(buildStatus: 'SUCCESSFUL')    // Notifies the Bitbucket instance of an SUCCESSFUL build
     } catch(err) {
         // Do clean up
-        notifyBitbucket(buildState: 'FAILED')        // Notifies the Bitbucket instance of an FAILED build
+        notifyBitbucket(buildStatus: 'FAILED')        // Notifies the Bitbucket instance of an FAILED build
     }
 ```
 
@@ -120,7 +120,7 @@ def notifyBitbucket(String state) {
             disableInprogressNotification: false,
             considerUnstableAsSuccess: true,
             ignoreUnverifiedSSLPeer: true,
-            buildState: state,
+            buildStatus: state,
             buildName: 'Performance Testing',
             includeBuildNumberInKey: false,
             prependParentProjectKey: false,
--- a/readme.md
+++ b/readme.md
@@ -82,6 +82,22 @@ node {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


scaytrase:
@westarne did you try these changes by yourself? I hope with current approach if you run notification multiple times per build than you get single build result in bitbucket, since they will override each other

westarne:
I have not actually imported this into Jenkins to run, yet. I think what you mean is that multiple notifications like `notifyBitbucket(buildName: "Build")` and `notifyBitbucket(buildName: "Test")` will overwrite each other. As far as I know, this is in fact the case, but you could use `notifyBitbucket(buildName: "Build", projectKey: "build")` to change this key. Maybe an option like `includeBuildNameInKey` would be good to do this automatically though.

scaytrase:
I think you can automatically append overriden build name to the project key if any.

westarne:
Ok i will do this later today or maybe tomorrow. I think I will rename buildState to buildStatus, too.

westarne:
Ok, so now the buildName is appended to the key whenever it is set. I could also add this to the default build key method, so it is only used, when the projectKey is not specifically set. I actually think the latter is the better option to let the user overwrite this if he wants.

scaytrase:
Yea, good idea

westarne:
Ok, i've changed it now to be in the default build key only.

westarne:
I have now tested this on my Jenkins with this code snippet, to test whether the calls are working and doing what intended:

```groovy
        notifyBitbucket(buildStatus: 'INPROGRESS', buildName: 'MY NAME')
        notifyBitbucket(buildStatus: 'SUCCESSFUL', buildName: 'MY NAME')
        notifyBitbucket(buildStatus: 'FAILED', buildName: 'MY NAME')
        notifyBitbucket(buildStatus: 'INPROGRESS', buildName: 'Build')
        notifyBitbucket(buildStatus: 'SUCCESSFUL', buildName: 'Build')
        notifyBitbucket(buildStatus: 'INPROGRESS', buildName: 'Build')
        notifyBitbucket(buildStatus: 'INPROGRESS')
        notifyBitbucket(buildStatus: 'SUCCESSFUL')
```

Status in Bitbucket:
image

So this works as intended. The builds still link as before.

scaytrase:
ok, thank you.I'll try to get into it this week

пн, 12 авг. 2019 г., 20:19 Arne Westphal <notifications@github.com>:

yueyehm:
@westarne btw, is it convenience for you to help making the "**_description_**" as overridable too? Current logic is using the build description, but if we could override the description when we invoke notifyStash method, it would be great.

scaytrase:
@yueyehm I think this could be done as separater PR another issue.
@westarne sorry, I've got a bit busy these days, will release it this week

scaytrase:
thank you @westarne

westarne:
Thank you, too.
@yueyehm if I find the time I can make another pull request for that.

yueyehm:
@westarne thanks very much!

Issue #245, opened 2019-09-04
Missing required parameter: "buildName"


adrian-at-dmi:
Just updated to v1.20 and received build errors with the following message:
```
WorkflowScript: 36: Missing required parameter: "buildName" @ line 36, column 5.
   				notifyBitbucket commitSha1: "${env.GIT_COMMIT}", considerUnstableAsSuccess: false, credentialsId: '<some-guid>', disableInprogressNotification: false, ignoreUnverifiedSSLPeer: false, includeBuildNumberInKey: true, prependParentProjectKey: false, projectKey: '', stashServerBaseUrl: '<some-url-to-bitbuckt-server>'
```

Looks like #241 introduced an backwards-incompatible change even though the issue suggests that the "buildName" parameter should be optional. The parameter also does not appear in the snippet generator which makes it hard to know what it does, how it works and that it exists in the first place.

Downgrading to 1.19 and a Jenkins restart fixed the issue for now, so it's not super urgent. But I wanted to let you know, that the new overwritable build name feature probably needs some additional work.

scaytrase:
Yeah, sounds like a bug, @westarne do you have a possibility to look into this?

westarne:
I actually don't know why this does not work with the other jenkins syntax. It definitely works with the `notifyBitbucket(...)` syntax without specifying a buildName. I will check the source code if I have missed something but I'm most likely not able to check this on an actual Jenkins. Are there already other optional parameters in the problematic syntax? Something that I can use as orientation?

@adrian-at-dmi if you add a buildName does it throw the same error for buildStatus?

westarne:
I am sorry, I could not reproduce the problem. Using the following in v1.20 I had not error and the status was published correctly in Bitbucket:

```groovy
def revision = "<my-revision>"
notifyBitbucket commitSha1: revision, considerUnstableAsSuccess: false, credentialsId: '<some-guid>', disableInprogressNotification: false, ignoreUnverifiedSSLPeer: false, includeBuildNumberInKey: true, prependParentProjectKey: false, projectKey: '', stashServerBaseUrl: '<some-url-to-bitbuckt-server>'
```

Maybe you are using a different version of the pipeline plugins like Pipeline: Declarative or Pipeline: Groovy or something else? We actually have not up-to-date versions of a lot of plugins. Maybe something has changed there which would need adaption in this plugin or your version is even older.

adrian-at-dmi:
[...]

PR #246, opened 2019-10-02
Implement setters instead of constructor for optional parameters

westarne:
Concerning #245 

I was not able to reproduce the original issue, but comparing with other plugins this might do the trick. For now this was not tested by me on an actual Jenkins and I could only test if it works for me as it did before. It would be great if someone, for whom the v1.20 is failing, could test this build. (@adrian-at-dmi, @tschlue, @r4d1um)

If you are not able to build this plugin (and you trust me), here is a locally built .hpi:
[stashNotifier.zip](https://github.com/jenkinsci/stashnotifier-plugin/files/3681822/stashNotifier.zip)


westarne:
I have now tested this build on my Jenkins and it still works for me. So it does not seem to break anything at least. But I cannot verify whether this actually makes the parameters optional as intended, but it should...

scaytrase:
@westarne it does not break already configured instances? I'll try to test it on my jenkins instance on weekend

westarne:
I've tested this via manually replacing the old *.jpi in the plugins folder, so that worked without changing anything configured. I'm not sure if this would work via the usual install process as well, but I think it should.

scaytrase:
sounds good

proski:
I believe this change will make it possible to use `notifyBitbucket` without arguments in declarative pipelines. That would be very good! I can do it by surrounding that command with `script { ... }`, but it shouldn't be necessary. There are no required arguments if the URL and credentials are set up in the global settings.

However, I'm concerned that this PR appears to be changing more code than necessary. In particular, it should not be adding extra settings. It's better to keep the configuration to the minimum. Even some of the existing overrides don't make sense to me. For instance, "Override project key". Why would anyone do it? And this PR is adding "Build Status". Why would anyone use anything but the actual build result?

In any case, adding new user interface should be a separate PR, with a separate justification.

westarne:
Both project key and build status are useful and already implemented and merged from different previous issues.
They are not added in this PR.

~Override the project key is useful to push separate statuses for different build steps in the pipeline as well as the build status.~ (this is actually the buildName not projectKey i meant)

You can this way use the custom buildStatus for example to say that the build step was successful and the test step is in progress. By setting the buildStatus to succesful once you could otherwise never push a "in progress" anymore after that. But as i said, both were implemented before and not in this PR.

scaytrase:
Yes, that's true, override project key and build status were implemented before

src/main/resources/org/jenkinsci/plugins/stashNotifier/StashNotifier/config.jelly
```diff
@@ -5,8 +5,14 @@
     </f:entry>
     <f:entry title="${%Credentials}" field="credentialsId">
       <c:select/>
-     </f:entry>
-     <f:entry title="Commit SHA-1" field="commitSha1">
+    </f:entry>
+    <f:entry title="Commit SHA-1" field="commitSha1">
+      <f:textbox />
+    </f:entry>
+    <f:entry title="Build Name" field="buildName">
```
proski, line 13: Why is this field being added in this PR?

westarne: Because these were forgotten previously and might be also connected to the bug. And even if not, they are fixing the implementation for the fields of the initial issue.

darxriggs:
I just saw that there is a merge conflict to be resolved.

So let me ask what is the current state of this pull request and the next steps?

scaytrase:
> So let me ask what is the current state of this pull request and the next steps?

On #245 people report problem is gone when they use `script` wrapper, so I'm not sure we are fixing a problem here. But as an improvement this might be good

westarne:
I've resolved the conflicts.

In my opinion this change could fix the issue #245. But I was not able to really test this because I could not reproduce the initial issue. It'd great if someone could test this manually (e.g. with the provided .hpi) who is able to reproduce the initial issue.

In general this should not create additional issues either way even if it does not fix #245 but testing it in integration first is still important I think.

darxriggs:
@westarne Do you know `mvn hpi:run`? It is very handy to interactively try out the latest changes during development.

I suggest to run it, install the "Pipeline: Declarative" plugin and try a Jenkinsfile like this:

```
pipeline {
   agent any

   stages {
      stage('Testing') {
         steps {
            echo 'Some Message'
            notifyBitbucket(
                stashServerBaseUrl: 'https://localhost',
                projectKey: 'KEE',
                commitSha1: 'shaaaaaaaa',
                credentialsId: 'example-id',
                considerUnstableAsSuccess: false,
                disableInprogressNotification: false,
                ignoreUnverifiedSSLPeer: false,
                includeBuildNumberInKey: false,
                prependParentProjectKey: false,
            )
         }
      }
   }
}
```

You should be able to reproduce the issue with this pipeline on the current `master` or latest tag `1.20` state and also be able to verify in your branch that it is fixed.

westarne:
> Do you know `mvn hpi:run`? It is very handy to interactively try out the latest changes during development.

Thank you for this advice, I was able to reproduce the original issue with this. And it seems to be resolved with this pull request.
Because the bitbucket server I usually use is behind a corporate firewall I was not able to actually push a status yet, because some other configurations need to be done. But using the `release/1.x` state with `buildName: '', buildStatus: ''` I get the exact same output I get with this PR branch without these parameters specified (and I get the initial missing parameters problem of course when I do not set the params on the `release/1.x` branch).

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -181,105 +181,144 @@ public BuildStepMonitor getRequiredMonitorService() {
...
-        return disableInprogressNotification;
+    public String getStashServerBaseUrl() {
+        return stashServerBaseUrl;
     }
 
-    public boolean isConsiderUnstableAsSuccess() {
-        return considerUnstableAsSuccess;
+    @DataBoundSetter
+    public void setStashServerBaseUrl(String stashServerBaseUrl) {
+        this.stashServerBaseUrl = stashServerBaseUrl != null && stashServerBaseUrl.endsWith("/")
```
darxriggs: [StringUtils#stripEnd](https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html#stripEnd(java.lang.String,%20java.lang.String)) could be used.

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -181,105 +181,144 @@ public BuildStepMonitor getRequiredMonitorService() {
...
     public StashBuildState getBuildStatus() {
         return buildStatus;
     }
 
+    public void setBuildStatus(StashBuildState buildStatus) {
+        this.buildStatus = buildStatus;
+    }
+
+    @DataBoundSetter
+    public void setBuildStatus(String buildStatus) {
```
darxriggs, line 259: The complete method could be stripped down to:
```
try {
    this.buildStatus = StashBuildState.valueOf(buildStatus);
} catch (Exception e) {
    // ignore unknown or null values
}
```

westarne:
Implemented the proposed optimizations

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -181,105 +184,140 @@ public BuildStepMonitor getRequiredMonitorService() {
...
-        }
-        this.buildStatus = overwrittenBuildState;
-
-        this.buildName = buildName;
-        this.includeBuildNumberInKey = includeBuildNumberInKey;
-        this.projectKey = projectKey;
-        this.prependParentProjectKey = prependParentProjectKey;
-        this.disableInprogressNotification = disableInprogressNotification;
-        this.considerUnstableAsSuccess = considerUnstableAsSuccess;
+    StashNotifier(
```
darxriggs: I am thinking if the number of constructors can be reduced even more? Is this one really required?

westarne: I'm not sure about that, and neither am I that the test cases existing as well as the ones I can do locally cover all potential use cases. So I've kept it to be on the safe side.

darxriggs: But this is a new one and another one has been removed.

In other plugins it's often handled the way to keep all old constructors for backward compatibility and annotate them with `@Deprecated`. That's something the maintainers should decide.

westarne: Oh you're right. Then I'm not totally sure, what the constructor was meant for. I think it's just the old `DataBoundConstructor` without all of the optional parameters. So I could just move the logic in the no-argument constructor.

westarne: Was some time out of the country and noticed that I did not push the change before. Done that now.

coxalass:
Hello,

I guess this might also fix the following issue : https://github.com/jenkinsci/stashnotifier-plugin/issues/266

Is there anything preventing this PR from being merged ?

Kind regard,

jetersen (approved):
This is usually best practice for Jenkins plugins and follows JCasC recommendations as well.

scaytrase:
At the moment I think I can accept this PR, but this should be updated

scaytrase:
@westarne a lot of compilation errors

westarne:
> @westarne a lot of compilation errors

never merge on the git website, noted. Was just missing a `}` as far as I can tell. Added that

scaytrase (approved):
LGTM 👍

scaytrase:
Thanks @westarne . I'll try try to release this as soon as I can

scaytrase:
https://github.com/jenkinsci/stashnotifier-plugin/releases/tag/1.25

PR #250, opened 2020-01-04
Simplify setting "compatibleSinceVersion"

darxriggs:
The used property is available since maven-hpi-plugin 3.2.
See https://github.com/jenkinsci/maven-hpi-plugin/blob/maven-hpi-plugin-3.2/README.md.


scaytrase (approved):

scaytrase:
Thank you @darxriggs !

PR #251, opened 2020-01-05
Enable InjectedTest & fix found issues

darxriggs:
- enable maven-enforcer-plugin to show plugin incompatibilities & fix them
  - without these changes the plugin would fail to start in InjectedTest
  - "git client" plugin v2.0.0 requires "structs" plugin v1.3 or later
    (it was a transitive dependency via "display-url-api" plugin v1.1.1)
  - "git" plugin v3.0.0 requires "credentials" plugin v2.1.4 or later
- all .jelly files should escape content per default
  - this was reported via InjectedTest
  - also see https://wiki.jenkins.io/display/JENKINS/Jelly+and+XSS+prevention
- explicit Hamcrest dependency declaration
  - Hamcrest is used in the tests and therefore should be declared explicitly
  - use the highest version to resolve version conflicts due to transisitivy

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>display-url-api</artifactId>
-            <version>1.1.1</version>
+            <version>2.0</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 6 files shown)


scaytrase (approved):

scaytrase:
Thanks @darxriggs !

PR #252, opened 2020-01-05
Use a mocked version of JenkinsLocationConfiguration in tests

darxriggs:
Otherwise its constructor calls `load()` which results in a `NullPointerException` because it tries to access the `ServletContext` which is not available in tests.

This exception did not affect the test results but was confusing when looking at the test outputs.


scaytrase (approved):

scaytrase:
Thanks @darxriggs !

PR #253, opened 2020-01-05
Simplify exception declaration in tests

darxriggs:
In tests it is fine and much easier to maintain to only use `throws Exception` on test methods.

On some tests it was already declared like this. Therefore it is consistent now.


scaytrase (approved):

scaytrase:
Thanks @darxriggs !

PR #256, opened 2020-01-05
Inline "lookupCredentials" methods

darxriggs:
There was a mix of the `lookupCredentials` wrapper methods and using `CredentialsProvider#lookupCredentials` directly. This also resulted in different ways of mocking these calls in the tests.

As the mentioned methods were originally introduced for mocking purposes and PowerMockito now is anyway used in the tests to mock static methods, there is no need for these wrapper methods anymore.


scaytrase:
Thanks @darxriggs !

PR #259, opened 2020-01-06
Use JUnit & PowerMock versions as defined in parent pom

darxriggs:
JUnit is also 4.12 in the parent pom. Therefore no changes regarding this.

PowerMock is 2.0.4 in the parent pom, which means an upgrade and required changes.

1) For simplicity of importing always use `org.hamcrest.Matchers` to import the Hamcrest matchers.

2) Use the new `org.mockito.ArgumentMatchers` instead of the old deprecated `org.mockito.Mockito` to import Mockito argument matchers.

3) Do not use the now deprecated `anyObject` Mockito argument matcher anymore.

4) The new Mockito argument matchers do not match `null` anymore. Therefore an additional `TokenMakro` mocking was required.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -109,19 +109,16 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.powermock</groupId>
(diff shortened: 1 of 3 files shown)


scaytrase (approved):

scaytrase:
Thank you @darxriggs !

PR #267, opened 2020-11-11
Fix build state examples in README

jojo221119:
Change BuildState 'SUCCESS' to 'SUCCESSFUL'

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -73,7 +73,7 @@ node {
     try {
         // Do stuff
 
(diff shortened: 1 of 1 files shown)


scaytrase:
Not sure, whether it's correct value. There is no issues about this before and code operates `SUCCESS` value in terms of Jenkins build state

https://github.com/jenkinsci/stashnotifier-plugin/blob/e1451493a13009bdbceb4710988ce6aacc9c251f/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java#L319

PR #303, opened 2022-04-01
added buildUrl property for overriding the url used for the build / run

### LINKED ISSUE #302: Allow overriding the build url used for the build
### What feature do you want to see added?

Currently the build url sent to bitbucket represents the top-level build, which works for most cases. In a pipeline build, you may want to notify bitbucket for individual stages, by setting the `buildName` prop for the stage. Jenkins allows you to view the logs for an individual stage, which is useful when you're trying to find what tests failed in a particular stage (url looks something like `job/PR-5339/19/execution/node/71/log/`). 

Coupling `buildName` with a new prop `buildUrl` would facilitate this. 

### Upstream changes

_No response_

marchaos:
This pull request allows overriding the url used for the build / run. This is useful in pipeline builds where you might notify bitbucket about a particular stage, where the url should go directly to the logs of that stage, rather than the top level pipeline. A new prop `buildUrl` allows you to override what is used, otherwise the default url that comes from `DisplayURLProvider` is used. 

See feature request for more info - https://github.com/jenkinsci/stashnotifier-plugin/issues/302

- [ x ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ x ] Ensure that the pull request title represents the desired changelog entry
- [ x ] Please describe what you did
- [ x ] Link to relevant issues in GitHub or Jira
- [ x ] Link to relevant pull requests, esp. upstream and downstream changes
- [ x ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -136,6 +136,12 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
     private String buildName;
 
     /**
+     * specify a build url to be included in the Bitbucket notification.
+     * If null, the usual url of the build will be used.
+     */
+    private String buildUrl;
+
+    /**
      * if true, the build number is included in the Bitbucket notification.
      */
     private boolean includeBuildNumberInKey;
@@ -182,6 +188,7 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
             String commitSha1,
             String buildStatus,
             String buildName,
+            String buildUrl,
             boolean includeBuildNumberInKey,
             String projectKey,
             boolean prependParentProjectKey,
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -1091,6 +1108,21 @@ protected String getBuildName(final Run<?, ?> run) {
...
 
+    /**
+     * Returns the build url to be pushed. This will select the specifically overwritten build url
+     * or get the build url from the {@link DisplayURLProvider}.
+     *
+     * @param run the run to notify Bitbucket of
+     * @return the url of the run
+     */
+    protected String getBuildUrl(final Run<?, ?> run) {
+        if (buildUrl != null && buildUrl.trim().length() > 0) {
```
offa: You can test for empty directly:
```suggestion
        if (buildUrl != null && !buildUrl.trim().isEmpty()) {
```

scaytrase (approved):
LGTM 👍 

sorry for long delay

PR #343, opened 2023-07-28
Fix "Invalid cookie header" warnings

offa:
Each status update sent to Bitbucket issues a warning in the Jenkins log:

> Invalid cookie header: "Set-Cookie: […]". Invalid 'expires' attribute: […]

This PR fixes the issue.

### Testing done

Verified on a Jenkins instance.

```[tasklist]
### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
+++ b/src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
@@ -50,6 +50,7 @@ import org.apache.http.HttpHeaders;
 import org.apache.http.HttpHost;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.AuthenticationException;
+import org.apache.http.client.config.CookieSpecs;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.config.Registry;
@@ -530,7 +531,8 @@ public class StashNotifier extends Notifier implements SimpleBuildStep {
         RequestConfig.Builder requestBuilder = RequestConfig.custom()
                                             .setSocketTimeout(timeoutInMilliseconds)
                                             .setConnectTimeout(timeoutInMilliseconds)
-                                            .setConnectionRequestTimeout(timeoutInMilliseconds);
+                                            .setConnectionRequestTimeout(timeoutInMilliseconds)
+                                            .setCookieSpec(CookieSpecs.STANDARD);
 
         HttpClientBuilder clientBuilder = HttpClients.custom();
         clientBuilder.setDefaultRequestConfig(requestBuilder.build());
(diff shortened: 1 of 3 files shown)


offa:
@sghill could you have a look at this? :-)

sghill:
@offa is it possible to add a test that verifies this behavior? MockServer or similar may be useful here

offa:
Thanks, I'll have a look at it!

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
```diff
@@ -530,7 +531,8 @@ protected CloseableHttpClient getHttpClient(PrintStream logger, Run<?, ?> run, S
         RequestConfig.Builder requestBuilder = RequestConfig.custom()
                                             .setSocketTimeout(timeoutInMilliseconds)
                                             .setConnectTimeout(timeoutInMilliseconds)
-                                            .setConnectionRequestTimeout(timeoutInMilliseconds);
+                                            .setConnectionRequestTimeout(timeoutInMilliseconds)
```
offa, line 534: According to usage search, the function isn't used anymore and deprecated. Thus, instead of adding tests to these changes, I can drop them. What do you think?

offa:
@sghill Test added (for the not-deprecated version; see comments above).

src/main/java/org/jenkinsci/plugins/stashNotifier/StashNotifier.java
sghill: It seems low risk so I think we can leave it. If there is an issue we can revert.

src/test/java/org/jenkinsci/plugins/stashNotifier/DefaultApacheHttpNotifierTest.java
```diff
@@ -154,4 +166,18 @@ public void notifyStash_fail() throws Exception {
         NotificationResult notificationResult = notifyStash(400);
         assertThat(notificationResult.indicatesSuccess, is(false));
     }
+
+    @Test
+    public void notifyStashUsesRequestParameters() throws Exception {
```
sghill, line 171: Thanks! I was hoping to get a test that demonstrates the server side cookie warning goes away with this setting, but I recognize that may be quite difficult to put together.

This approach at least clarifies it's intentional to set this configuration.

sghill (approved):

Issue #290, opened 2022-02-07
Bitbucket 7.15 RestAPI update


swaragade:
### What feature do you want to see added?

StashNotifier plugin (v1.25) uses a Bitbucket API [here ](https://github.com/jenkinsci/stashnotifier-plugin/blob/release/1.x/src/main/java/org/jenkinsci/plugins/stashNotifier/BuildStatusUriFactory.java#L13) for sending the build notifications.
This API has been deprecated by the Bitbucket.
Please add support for [new BitBucket Notification API](https://docs.atlassian.com/bitbucket-server/rest/7.15.1/bitbucket-rest.html#idp226)

swaragade:
@scaytrase kindly check

scaytrase:
@swaragade currently I have neither bitbucket instance nor active jenkins installations, so It would be hard to implement this feature by myself. But any PR is welcome here

offa:
@swaragade are you sure? The endpoint is still listed without any deprecation notice: https://developer.atlassian.com/server/bitbucket/how-tos/updating-build-status-for-commits/

roman-mibex-2:
This API is deprecated. However, it wasn't marked properly. https://jira.atlassian.com/browse/BSERV-13381 as such.

I got marked properly as deprecated in 8.3: https://developer.atlassian.com/server/bitbucket/rest/v803/api-group-build-status/#api-build-status-latest-commits-commitid-get

roman-mibex-2:
The new API requires the projectKey and repoSlug info. This isn't present in the plugin as it is designed now.
The projectKey can be supplied optionally.

So, using the new API would require some changes in the design or usage. That would need way more though and time than I can offer at this point in time.

brucewilsn:
Updating to the new API would be more than just nice at this point. The old API is partly broken. Bitbucket will show the build status next to a commit. But even if the source branch tip shows a successful build, it will not allow merging a pull request (if the target branch is configured to require a successful build.)

PR #361, opened 2023-10-03
Issue #290: Implement compatibility with new Bitbucket rest api

BUGinator86:
and fix missing "parent" property in JSON for required builds.

This fixes Issue #290 without breaking compatibility with older versions of BitBucket/Stash.
We support the old deprecated REST API and the new REST API.

### Testing done

The test was done in our test environment with BitBucket 7.21 that supports both versions of the API, because we needed a slim plugin for our required builds.
After both versions successfully worked in our test environment we now use it for about 3 Weeks in production without any problems.

```[tasklist]
### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
                 <artifactId>bom-2.426.x</artifactId>
-                <version>3056.v53343b_a_b_a_850</version>
+                <version>3080.vfa_b_e4a_a_39b_44</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>


src/main/java/org/jenkinsci/plugins/stashNotifier/BuildStatusUriFactory.java
```diff
@@ -8,9 +8,16 @@ public class BuildStatusUriFactory {
...
+    public static URI create(String baseUri, String projectKey, String slug, String commit) {
         String tidyBase = StringUtils.removeEnd(baseUri, "/");
-        String uri = String.join("/", tidyBase, "rest/build-status/1.0/commits", commit);
+
+        String uri;
+        if(projectKey == null || slug == null || projectKey.isEmpty() || slug.isEmpty()) {
+            uri = String.join("/", tidyBase, "rest/build-status/1.0/commits", commit);
+        }
+        else {
+            uri = String.join("/", tidyBase, "rest/api/latest/projects", projectKey, "repos", slug, "commits", commit, "builds");
```
gruetter: The implementation looks fine to me. However, the tests currently cover the legacy API calls (w/o project key and repo slug), only. I suggest you add another test for the new API, as well.

gruetter (approved):
This PR looks good to me and is rather relevant for users of this plugin. I suggest to add one more test case (see my comments below). 

@scaytrase and @sghill: My Java skills have become somewhat rusty (no pun intended) in the past years. I would appreciate either one of you give the code a looksee, as well, before merge.

sghill:
Seconding @gruetter's request for more test coverage.

I think we should also consider whether this makes sense to add as a separate notifier rather than modifying the existing one. There is support within the plugin for changing the notifier by system property, which would mean any issues with the new notifier could be easily worked around. Would love to hear more feedback about this idea.

PR #391, opened 2024-05-30
Enhancement/restapi update

BUGinator86:
Updated Pullrequest to also include tests for requests with bitbucket projectkey and slug.
Also put creation of payload in HttpNotifier classes to make it possible to create different payloads for different versions of stash/bitbucket.
Introduces ExtendedApacheHttpNotifier for extended payload and new selector that creates the ExtendedApacheHttpNotifier only if a BitBucket projectkey and slug is provided.
Else the behaviour is the same as before.

### Submitter checklist
- [x ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x ] Ensure that the pull request title represents the desired changelog entry
- [x ] Please describe what you did
- [x ] Link to relevant issues in GitHub or Jira
- [x ] Link to relevant pull requests, esp. upstream and downstream changes
- [x ] Ensure you have provided tests - that demonstrates feature works or fixes the issue


src/main/java/org/jenkinsci/plugins/stashNotifier/Notifiers/DefaultApacheHttpNotifier.java
```diff
@@ -51,32 +52,34 @@
...
             HttpResponse res = client.execute(req);
+
             if (res.getStatusLine().getStatusCode() != 204) {
                 return NotificationResult.newFailure(EntityUtils.toString(res.getEntity()));
             } else {
                 return NotificationResult.newSuccess();
             }
         } catch (Exception e) {
-            LOGGER.warn("{} failed to send {} to Bitbucket Server at {}", context.getRunId(), payload, uri, e);
+            LOGGER.warn("{} failed to send {} to Bitbucket Server at {}", context.getRunId(), payload, uri);
```
sghill, line 75: Why drop logging the exception here?

src/main/java/org/jenkinsci/plugins/stashNotifier/Notifiers/DefaultApacheHttpNotifier.java
```diff
@@ -94,7 +97,7 @@ else if (credentials instanceof StringCredentials) {
                 req.addHeader(HttpHeaders.AUTHORIZATION, "Bearer " + ((StringCredentials)credentials).getSecret().getPlainText());
             } 
             else {
-                throw new AuthenticationException("Unsupported credials");
+                throw new AuthenticationException("Unsupported credentials");
```
sghill, line 100: Nice, appreciate this update 👍

src/main/java/org/jenkinsci/plugins/stashNotifier/NotifierSelectors/HttpNotifierSelector.java
```diff
@@ -1,4 +1,4 @@
-package org.jenkinsci.plugins.stashNotifier;
+package org.jenkinsci.plugins.stashNotifier.NotifierSelectors;
```
sghill, line 1: I appreciate the effort to keep similar implementations together, but unfortunately this is a breaking change.

Can we leave it in the original package to minimize the work consumers will have to do?

src/main/java/org/jenkinsci/plugins/stashNotifier/BuildStatusUriFactory.java
```diff
@@ -8,9 +8,16 @@ public class BuildStatusUriFactory {
     private BuildStatusUriFactory() {
     }
 
-    public static URI create(String baseUri, String commit) {
```
sghill, line 11: Can we leave this method around and create a new overload for the extra parameters?

src/main/java/org/jenkinsci/plugins/stashNotifier/NotificationContext.java
```diff
@@ -10,10 +10,12 @@
 public class NotificationContext {
     private final PrintStream logger;
     private final String runId;
+    private final BuildInformation buildInformation;
 
-    public NotificationContext(PrintStream logger, String runId) {
```
sghill, line 14: Can we leave the old constructor too? Happy to mark it deprecated, just would prefer the next release doesn't cause compilation issues if possible.

src/main/java/org/jenkinsci/plugins/stashNotifier/NotifierSelectors/DefaultHttpNotifierSelector.java
```diff
@@ -0,0 +1,37 @@
...
+import org.jenkinsci.plugins.stashNotifier.Notifiers.ExtendedApacheHttpNotifier;
+import org.jenkinsci.plugins.stashNotifier.Notifiers.HttpNotifier;
+import org.jenkinsci.plugins.stashNotifier.SelectionContext;
+
+import java.util.*;
+
+public class DefaultHttpNotifierSelector implements HttpNotifierSelector {
+    private final List<HttpNotifier> httpNotifiers;
+
+    public DefaultHttpNotifierSelector(List<HttpNotifier> notifiers) {
```
sghill, line 14: Are there more than two implementations?

I like the idea here a lot. I think it would be quite a bit simpler by having the constructor take the `Default` and `Extended` notifiers directly, rather than iterating a list for a known value.

src/main/java/org/jenkinsci/plugins/stashNotifier/NotifierSelectors/HttpNotifierSelector.java
```diff
@@ -21,10 +24,10 @@ public interface HttpNotifierSelector {
...
      *
-     * @see StashNotifier#prebuild(AbstractBuild, BuildListener)
-     * @see StashNotifier#perform(Run, FilePath, Launcher, TaskListener)
      * @param context parameters useful for selecting a notifier
      * @return selected notifier
+     * @see StashNotifier#prebuild(AbstractBuild, BuildListener)
+     * @see StashNotifier#perform(Run, FilePath, Launcher, TaskListener)
      */
-    @NonNull HttpNotifier select(@NonNull SelectionContext context);
+    HttpNotifier select(@NonNull SelectionContext context);
```
sghill, line 32: Can we bring back the `NonNull` declaration?

src/main/java/org/jenkinsci/plugins/stashNotifier/Notifiers/DefaultApacheHttpNotifier.java
```diff
@@ -51,32 +52,34 @@
 import java.security.NoSuchAlgorithmException;
 import java.security.UnrecoverableKeyException;
 
-class DefaultApacheHttpNotifier implements HttpNotifier {
+public class DefaultApacheHttpNotifier implements HttpNotifier {
+    protected static final int MAX_FIELD_LENGTH = 255;
+    protected static final int MAX_URL_FIELD_LENGTH = 450;
```
sghill, line 57: Nice to see these additional validations 👍

src/main/java/org/jenkinsci/plugins/stashNotifier/Notifiers/DefaultApacheHttpNotifier.java
```diff
@@ -51,32 +52,34 @@
 import java.security.NoSuchAlgorithmException;
 import java.security.UnrecoverableKeyException;
 
-class DefaultApacheHttpNotifier implements HttpNotifier {
+public class DefaultApacheHttpNotifier implements HttpNotifier {
```
sghill, line 55: What do you think of adding a `DelegatingHttpNotifier` that makes the decision of which notifier to use based on the context? I've found that a simple pattern that makes it easy to retain API compatibility.

```
class DelegatingHttpNotifier implements HttpNotifier {
    private final HttpNotifier standard;
    private final HttpNotifier extended;

    NotificationResult send (context, ...) {
        return context.supportsExtended() ? extended.send(...) : standard.send(...)
    } 
}
```

sghill (changes requested):
Thank you for the effort in testing and designing something that works on multiple versions of Bitbucket! I added a few comments around retaining API compatibility so the latest update won't break compilation for folks, but other than that I think the direction is really solid.

antonakospanos:
Can we proceed with that? Using the new API is required so that we can pass the `parent` field too which is the one that Bitbucket searches to match with the newly introduced `Required Builds` before a PR can be merged.

